PR #97, opened 2014-12-31
Cached builds

samertm:
Source units now point to their previous revision if the unit isn't changed, and those units aren't rebuilt during the `make` step.

### DIFF
--- a/buildstore/store.go
+++ b/buildstore/store.go
@@ -1,7 +1,6 @@
 package buildstore
 
 import (
-	"fmt"
 	"log"
 	"os"
 	"path/filepath"
@@ -42,8 +41,6 @@ type RepoBuildStore interface {
 	// specific commit.
 	Commit(commitID string) rwvfs.WalkableFileSystem
 
-	Symlink(oldname, newname string) error
-
 	// FilePath returns the path (from the repo build store's root) to
 	// a file at the specified commit ID.
 	FilePath(commitID string, file string) string
@@ -72,13 +69,6 @@ func LocalRepo(repoDir string) (RepoBuildStore, error) {
 
 type repoBuildStore struct {
 	fs rwvfs.WalkableFileSystem
-}
-
-func (s *repoBuildStore) Symlink(oldname, newname string) error {
-	if fs, ok := s.fs.(rwvfs.LinkFS); ok {
-		return fs.Symlink(oldname, newname)
-	}
-	return fmt.Errorf("this file system does not support symlinks")
 }
 
 func (s *repoBuildStore) Commit(commitID string) rwvfs.WalkableFileSystem {
--- a/buildstore/store.go
+++ b/buildstore/store.go
@@ -1,6 +1,7 @@
 package buildstore
 
 import (
+	"fmt"
 	"log"
 	"os"
 	"path/filepath"
@@ -41,6 +42,8 @@ type RepoBuildStore interface {
 	// specific commit.
 	Commit(commitID string) rwvfs.WalkableFileSystem
 
+	Symlink(oldname, newname string) error
+
 	// FilePath returns the path (from the repo build store's root) to
 	// a file at the specified commit ID.
 	FilePath(commitID string, file string) string
@@ -69,6 +72,13 @@ func LocalRepo(repoDir string) (RepoBuildStore, error) {
 
 type repoBuildStore struct {
 	fs rwvfs.WalkableFileSystem
+}
+
+func (s *repoBuildStore) Symlink(oldname, newname string) error {
+	if fs, ok := s.fs.(rwvfs.LinkFS); ok {
+		return fs.Symlink(oldname, newname)
+	}
+	return fmt.Errorf("this file system does not support symlinks")
 }
 
 func (s *repoBuildStore) Commit(commitID string) rwvfs.WalkableFileSystem {
--- a/src/config_cmd.go
+++ b/src/config_cmd.go
@@ -132,21 +132,6 @@ func (c *ConfigCmd) Execute(args []string) error {
 	commitFS := buildStore.Commit(localRepo.CommitID)
 
 	// Write source units to build cache.
-	//
-	// TODO(sqs): create Makefile.config that makes it more standard to recreate
-	// these when the source unit defns change (and to determine when the source
-	// unit defns change), with targets like:
-	//
-	// UNITNAME/UNITTYPE.unit.v0.json: setup.py mylib/foo.py
-	//   src config --unit=UNITNAME@UNITTYPE
-	//
-	// or maybe a custom stale checker is better than just using file mtimes for
-	// all the files (maybe just use setup.py as a prereq? but then how will we
-	// update SourceUnit.Files list? SourceUnit.Globs could help here...)
-	//
-	// note(samertm): going with a custom checker because we need to create a
-	// symlink to the previous unit/graph files if a source unit hasn't
-	// changed. Punting on updating SourceUnit.Files for now.
 	if !c.NoCacheWrite {
 		if err := rwvfs.MkdirAll(commitFS, "."); err != nil {
 			return err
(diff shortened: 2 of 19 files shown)


src/repo_config.go
```diff
@@ -95,6 +95,32 @@ func resolveWorkingTreeRevision(vcsType string, dir string) (string, error) {
...
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		return nil, err
+	}
+	return strings.Split(string(bytes.TrimSpace(out)), "\n"), nil
+}
+
+func changedFilesFromCurrentRev(vcsType string, dir string, fromRev string) ([]string, error) {
+	if vcsType != "git" {
+		return nil, fmt.Errorf("changedFilesFromCurrentRev: unsupported vcs type: %q", vcsType)
```
samertm: If this errors, srclib prints a warning and builds the repository
normally.

Quinn Slack writes:

>> +func listRevisions(vcsType string, dir string) ([]string, error) {
>> +	if vcsType != "git" {
>> +		return nil, fmt.Errorf("listRevisions: unsupported vcs type: %q", vcsType)
>> +	}
>> +	cmd := exec.Command("git", "rev-list", "--max-count=5", "HEAD")
>> +	cmd.Dir = dir
>> +	out, err := cmd.CombinedOutput()
>> +	if err != nil {
>> +		return nil, err
>> +	}
>> +	return strings.Split(string(bytes.TrimSpace(out)), "\n"), nil
>> +}
>> +
>> +func changedFilesFromCurrentRev(vcsType string, dir string, fromRev string) ([]string, error) {
>> +	if vcsType != "git" {
>> +		return nil, fmt.Errorf("changedFilesFromCurrentRev: unsupported vcs type: %q", vcsType)
>
> It's fine to not implement this (and `listRevisions`) for hg, but does this mean that all builds of hg repos will now fail? Or is the error ignored? It should revert to the old behavior for unsupported VCS types (hg).
>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/sourcegraph/srclib/pull/97/files#r22396120

-- 
    Samer

src/repo_config.go
```diff
@@ -95,6 +95,32 @@ func resolveWorkingTreeRevision(vcsType string, dir string) (string, error) {
 	return strings.TrimSuffix(string(bytes.TrimSpace(out)), "+"), nil
 }
 
+func listRevisions(vcsType string, dir string) ([]string, error) {
```
samertm: Haha, I had read something about Rob Pike (pbuh) preferring `(arg0 type0, arg1
type0)` to `(arg0, arg1 type0)` because of how it looks in godoc, but
the standard library has many more examples of the latter :)


Quinn Slack writes:

>> @@ -95,6 +95,32 @@ func resolveWorkingTreeRevision(vcsType string, dir string) (string, error) {
>>  	return strings.TrimSuffix(string(bytes.TrimSpace(out)), "+"), nil
>>  }
>>  
>> +func listRevisions(vcsType string, dir string) ([]string, error) {
>
> `vcsType, dir string`
>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/sourcegraph/srclib/pull/97/files#r22396127

-- 
    Samer

sqs: Oh. If it's intentional it's fine. I just thought you had unintentionally done it.

src/repo_config.go
```diff
@@ -95,6 +95,32 @@ func resolveWorkingTreeRevision(vcsType string, dir string) (string, error) {
...
+	cmd := exec.Command("git", "rev-list", "--max-count=5", "HEAD")
+	cmd.Dir = dir
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		return nil, err
+	}
+	return strings.Split(string(bytes.TrimSpace(out)), "\n"), nil
+}
+
+func changedFilesFromCurrentRev(vcsType string, dir string, fromRev string) ([]string, error) {
```
samertm: Just to clarify, changedFilesFromCurrentRev only returns the files that
have changed between fromRev and HEAD. I hadn't thought of the case
where srclib is building uncommitted code, and I think
changedFilesFromCurrentRev should diff with the current working
directory to be consistent with srclib's current behavior.

Let me fix up the name, though.

Quinn Slack writes:

>> @@ -95,6 +95,32 @@ func resolveWorkingTreeRevision(vcsType string, dir string) (string, error) {
>>  	return strings.TrimSuffix(string(bytes.TrimSpace(out)), "+"), nil
>>  }
>>  
>> +func listRevisions(vcsType string, dir string) ([]string, error) {
>> +	if vcsType != "git" {
>> +		return nil, fmt.Errorf("listRevisions: unsupported vcs type: %q", vcsType)
>> +	}
>> +	cmd := exec.Command("git", "rev-list", "--max-count=5", "HEAD")
>> +	cmd.Dir = dir
>> +	out, err := cmd.CombinedOutput()
>> +	if err != nil {
>> +		return nil, err
>> +	}
>> +	return strings.Split(string(bytes.TrimSpace(out)), "\n"), nil
>> +}
>> +
>> +func changedFilesFromCurrentRev(vcsType string, dir string, fromRev string) ([]string, error) {
>
> "Current revision" is ambiguous. It could imply that it finds the files that changed between fromRev and the HEAD commit. But this command also gets files that have changed in the index (i.e., files that haven't yet been committed). The existing behavior is a bit weird, in that it'll process the index and store it in the `.srclib-cache/COMMITID` dir for the commit, but we should try to be explicit about that. So, what about calling this func `filesChangedFromRevToHEAD`?
>
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/sourcegraph/srclib/pull/97/files#r22396171

-- 
    Samer

sqs: Doesn't HEAD include the index and working tree?

sqs: Oh Nevermind I am wrong

unit/source_unit.go
```diff
@@ -77,12 +77,33 @@ type SourceUnit struct {
...
+	CachedRev string
+
 	// TODO(sqs): add a way to specify the toolchains and tools to use for
 	// various tasks on this source unit
 }
 
 //END SourceUnit OMIT
 
+// Doesn't do globs.
+func (u SourceUnit) ContainsAny(filenames []string) bool {
```
samertm: This changeset doesn't handle deleted files. If a file is deleted from
a source unit, the source unit will only be recomputed if another file
that belongs to the source unit is changed. It makes sense to fix this
eventually, but I didn't think it was a blocking issue.

It will handle renamed or moved files if they're added to git's index.

The trade-off here is whether to rely on the file system's state or
the scm's state when making the decision of whether we should cache a
build file. I decided to go with the scm because there's an action
associated with changed and unchanged source units: changed source
units need to be rebuilt, and unchanged source units need to be
copied. Relying on a makefile makes sense when the choice is between
action and inaction. Additionally, we want srclib to know what
source units have changed when we start rebuilding unchanged source
units that depend on units that have changed, and to check if a cached
build file exists before copying it. I think that relying on the scm
for this information is more clear.

If we use globs to see if a makefile has changed, then we're relying
on every toolchain to pass us the correct globs (which isn't a big
deal, but I think that doing less work with each individual toolchain
is better reliability's sake), and we're using the state of the
filesystem to tell us whether a file changed, instead of the scm. I
don't think that's a big deal for builds on our site, because the
behavior will be the same using either method, but for using srclib
interactively, "srclib builds every unit that changed since the last
built commit" is more clear to me than "srclib builds every unit since
the last time it ran". If a build of source unit can fail based on the
state of another source unit, then it's less reliable to rebuild based
on the file's modification time.

I think getting the state of a repo from the scm makes more sense, but
I'm open to the idea of using globs and mtimes instead if there's a
convincing case for it.

This branch will do the right thing when given your example if the new
file is added to git's index (once I finish responding to your comments).

src/config_cmd.go
```diff
@@ -142,15 +142,47 @@ func (c *ConfigCmd) Execute(args []string) error {
 	// or maybe a custom stale checker is better than just using file mtimes for
 	// all the files (maybe just use setup.py as a prereq? but then how will we
 	// update SourceUnit.Files list? SourceUnit.Globs could help here...)
+	//
+	// note(samertm): going with a custom checker because we need to create a
```
samertm: By "custom checker", I mean we're relying on srclib & the scm for whether we should rebuild a unit, not globs and the makefile. This comment is confusing, I'll remove it.

buildstore/store.go
```diff
@@ -71,6 +74,13 @@ type repoBuildStore struct {
 	fs rwvfs.WalkableFileSystem
 }
 
+func (s *repoBuildStore) Symlink(oldname, newname string) error {
+	if fs, ok := s.fs.(rwvfs.LinkFS); ok {
+		return fs.Symlink(oldname, newname)
+	}
+	return fmt.Errorf("this file system does not support symlinks")
```
samertm: I don't use it consistently, I'll add it to my save hook.

samertm:
The test errors out because of errors in fsouza/go-dockerclient, but they pass locally.

PR #107, opened 2015-01-19
Add raw keys for unresolved resolutions

xizhao:
Currently unresolved resolutions have empty targets and break keyid()


samertm:
Sorry for taking ages with this! This is a good change, merged here: https://github.com/sourcegraph/srclib/commit/5f5a0909d0e49e265b2b937e5ae7224e6f87fe40

PR #123, opened 2015-03-11
Interactive

samertm:
# Overview

`src i` is a readline-like interface for interacting with the structure of a project's source code

# Language

`src i`'s language is lightweight and inspired by SQL. The top priority for the language is **consistency**. All features exposed by `src i` should be consistent with the language as a whole.

## Examples
`src>` is the prompt.
### Simple Query
`src> SampleFunc`
 - Returns the definition declaration, body, and refs where the definition name starts with "SampleFunc".

The above command has some keyword modifiers implicitly set. It is equivalent to the following explicit form:
`src> :def SampleFunc :in . :format decl, body :select defs, refs`

#### Output

```
src> SampleFunc
---------- def ----------
func SampleFunc() SampleType
main.go:8:func SampleFunc() SampleType {
main.go:9-      return SampleType{1, 2}
main.go:10-}
---------- refs ----------
main.go:13:     SampleFunc()
main.go:18:     SampleFunc()
```

## Grammar

```
         <input> ::= { [ <keyword> ] <whitespace> <term-list> }
       <keyword> ::= ":" <keyword-part>
  <keyword-part> ::= defined by the list of keywords.
     <term-list> ::= <term> "," <term-list> | <term>
          <term> ::= <term-part>
                   | "'" { <term-part> } "'"
                   | "\"" { <term-part> } "\""
     <term-part> ::= <word> { <whitespace> [ <word> ] }
<opt-whitespace> ::= <whitespace> | ""
    <whitespace> ::= a string defined by the perl regexp ("\h" is the horizontal whitespace class): [\h]+
          <word> ::= a string defined by the perl regexp: [\w-()*_]+
```

TODO: add escaped quotes to the above.

## Keywords:
Normal keywords define the constraints of a query.

### `:def <names>`
Narrows defs to all defs matching `names`. If `names` is empty, then defs are not filtered by name.

`:def` is special because it is implicit if no keyword is used. For example, the input `some word` is equivalent to `:def some word`.
### `:in <repos>`
Changes the repository to `repo`.

Default value: repos = the current repo
### `:select <objects>`
Select only `objects`. Valid `objects` terms are: "defs", "refs", "docs".

Default value: `objects = defs, refs`
### `:format <args>`
Show defs in the format specificed by `args`. Valid `args` terms are: "decl", "methods", "body", "full".

Default value: `args = decl, body`
#### `:format "decl"`
Show def declarations.
#### `:format "methods"`
Show def methods.
#### `:format "body"`
Show the def body as it appears in the source.
#### `:format "full"`
Show all of the type information on the def. This includes the def's class heirarchy, type, methods.
### `:limit <number>`
Limit the results to `number`. `number` must be a valid number

## Special Keywords:
Special keywords are for interacting with `src i` in a meta sense, and for getting `src i`'s state. If any special keywords are used, a query is not run.

### `:help <topics>`
Show the help for `topics`. If `topics` is empty, show general help.

# TODO

## Bugs

- [ ] 'src config' won't overwrite an old config? or it doesn't delete old configs.
- [x] hide 'src config' and 'src make' output by default (or do something nice with it)
- [x] definition shows twice. build data wasn't overwritten? (src i)
- [ ] code more, grep less

## High priority

- [x] add lexer/parser
- [ ] implement all keywords
  - [x] :def
  - [ ] :in
  - [x] :select
  - [ ] :format
    - [x] decl
    - [x] body
    - [ ] methods
    - [ ] full
  - [x] :limit
  - [x] :help
- [x] hide build messages
- [ ] make src i permissive wrt where it's called:
  - [ ] work from the subdir of a project
  - [ ] work in projects that have no vcs (deal w/ ambiguities simply: if the working dir is nested in a git repo *at all*, that top-level git repo is considered the project. Otherwise, the project is seen from the current working directory.)
  - [ ] work in a project with vcs but no commits
  - [x] work in projects with no remotes
- [ ] figure out type queries

## Else

- [ ] move .srclib-(cache|store) to a central directory (similar to Emacs backups, like: ~/.srclib/res/!path!to!repo/.srclib-cache)
- [ ] finalize language, docs
  - [ ] add escaped quotes to grammar
- [ ] state what repository src i is connected to on start up
- [ ] completion
- [ ] color (might be too hard w/ liner)


sqs:
Initial feedback:

* [ ] Suppress the output during the build, and only show it if there's an error. (Show a loading screen in the meantime.) It's a lot of noise otherwise.
* [ ] Once the `src>` prompt is ready, explain what the user should do -- or tell her how to get help. My first session looked like:

```
src> ?
0:1: unexpected char: '[%!s(int32=63)]'
src> help

src> \h

src> \?

src> :help

src> :?
invalid keyword: keyword is empty
src> ^C
$ src i --help
Usage:
  src [OPTIONS] interactive

The interactive (i) command is a readline-like interface for interacting with the build data for a
project.

Global options:
  -v              show verbose output

Help Options:
  -h, --help      Show this help message
```

* [ ] Add tab-completion support. I typed "Con" and "Ope" expecting TAB to complete those to Config and Open, or similar, but pressing TAB has no effect. This might be caused by the next issue.
* [ ] The build appeared to have been successful (the build that was triggered by running `src i` in a repo with no existing build data for the current commit). But typing "Config" (the name of a defn in the srclib repo) and hitting ENTER doesn't yield any output other than a blank line. Did the build not succeed? If the build didn't succeed, or if it succeeded but no defs were found, it'd be good to tell the user.


(Then I came to this PR to see how to use it.)

PR #124, opened 2015-03-11
unit: don't omit empty `Files` slice when marshaling.

MaikuMori:
`sourcegraph/srclib-javascript` expects that unit json will always contain `Files` key.

Fixes sourcegraph/srclib-javascript#11.

This should have no side effects in other toolchains, but I didn't check.


sqs:
Thanks!

PR #125, opened 2015-03-12
toolchain: fix find failing when directory contains file symlink

MaikuMori:
Seems like I manage to use srclib in unorthodox ways.

Basically I figured why not set $SRCLIBPATH to something like `$HOME/.srclib:$GOPATH/src` (it as it turns out not a good idea anyway).
And one of the packages had a symlink to file which broke `toolchain.List`

Before fix it would fail with:

> lstat /some/something/README.md: not a directory

if `README.md` was a symlink.


samertm:
This looks good to me. Thanks for the PR!

PR #126, opened 2015-03-13
Return error on import failure

beyang:
Returns error if a build fails to import, instead of just logging it.


beyang:
@samertm is there any reason why it logs and doesn't return an actually error?

samertm:
No. This command isn't used on our site, so this change doesn't break anything.

PR #127, opened 2015-03-17
Add temp dir cmd

MaikuMori:
Should still decide on temp dir location. Using `~/.srclib/tmp` doesn't seem to be wise since `src` also looks for toolchains there.

### DIFF
--- a/src/toolchain_cmd.go
+++ b/src/toolchain_cmd.go
@@ -292,7 +292,7 @@ func (c *ToolchainAddCmd) Execute(args []string) error {
 
 type ToolchainTempDirCmd struct {
 	Args struct {
-		ToolchainPath string `name:"TOOLCHAIN" default:"." description:"toolchain path for which to get temp dir"`
+		ToolchainPath string `name:"TOOLCHAIN" description:"toolchain path for which to get temp dir"`
 	} `positional-args:"yes" required:"yes"`
 }
 
--- a/src/toolchain_cmd.go
+++ b/src/toolchain_cmd.go
@@ -73,6 +73,15 @@ func init() {
 		log.Fatal(err)
 	}
 
+	_, err = c.AddCommand("temp-dir",
+		"get toolchain's temp dir",
+		"Get toolchain's temp directory. Creates it if it doesn't exists.",
+		&toolchainTempDirCmd,
+	)
+	if err != nil {
+		log.Fatal(err)
+	}
+
 	_, err = c.AddCommand("install",
 		"install toolchains",
 		"Download and install toolchains",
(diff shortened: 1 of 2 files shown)


src/toolchain_cmd.go
```diff
@@ -73,6 +73,15 @@ func init() {
 		log.Fatal(err)
 	}
 
+	_, err = c.AddCommand("temp-dir",
+		"get toolchain's temp dir",
+		"Get toolchain's temp directory. Creates it if it doesn't exists.",
```
samertm: - [ ] I'm not sure what standard we're following for grammar, but these two lines should be "get a toolchain's temp dir" and "Get a toolchain's temporary directory. Create it if it doesn't exist."
The second line is seen as documentation, so we try not to use abbreviations like "temp".

src/toolchain_cmd.go
```diff
@@ -281,6 +290,24 @@ func (c *ToolchainAddCmd) Execute(args []string) error {
 	return toolchain.Add(c.Dir, c.Args.ToolchainPath)
 }
 
+type ToolchainTempDirCmd struct {
+	Args struct {
+		ToolchainPath string `name:"TOOLCHAIN" default:"." description:"toolchain path for which to get temp dir"`
```
samertm: - [ ] Because TOOLCHAIN is required, you should remove 'default:"."'

toolchain/temp_dir.go
```diff
@@ -0,0 +1,38 @@
...
+	"fmt"
+	"os"
+	"path/filepath"
+	"strings"
+
+	"sourcegraph.com/sourcegraph/srclib"
+)
+
+// TEMPDIRNAME is directory under SRCLIBPATH where to store temp directories for toolchains.
+const TEMPDIRNAME = "tmp"
```
samertm: - [ ] We don't use all-caps for constants (see https://sourcegraph.com/github.com/MaikuMori/srclib@540440047bb78d0fd3a0a3e6b295f67c61635fc9/.tree/buildstore/store.go#def=/github.com/MaikuMori/srclib@master/.GoPackage/github.com/MaikuMori/srclib/buildstore/.def/BuildDataDirName&startbyte=336&endbyte=370). TempDirName works.

samertm:
We really really need multiple levels of tests on travis, but this passes the go-level test. I can merge it after you address the above comments.

MaikuMori:
What about directory location?

samertm:
Oh yeah. It makes sense to special case the toolchain lookup function so that it doesn't look in "tmp" for toolchains.

samertm:
lgtm :+1: nice work

PR #147, opened 2015-05-21
Implement src push --commit

keegancsmith:
`src push` is a wrapper around `src build-data upload` and `src remote
import-build` both of which support the `--commit` flag. This commit implements
optionally specifying commit for `src push`, so one does not need to checkout
the wanted revision.

### DIFF
--- a/src/push_cmd.go
+++ b/src/push_cmd.go
@@ -20,6 +20,7 @@ func init() {
 }
 
 type PushCmd struct {
+	CommitID string `long:"commit" description:"commit ID of build data to operate on"`
 }
 
 var pushCmd PushCmd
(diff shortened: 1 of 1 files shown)


keegancsmith:
The implementation of this feels pretty hacky. Feedback welcome :)

sqs:
Looks good to me - what parts do you think are hacky?

(Tests pass if the go-sourcegraph dep is on the nodb branch, but Travis CI uses master.)

keegancsmith:
Having to set CommitID on each of the Opt structs, feels somewhat fragile for future updates. But otherwise, I'm happy with it

sqs:
Yeah, true. I think you could also set `localRepo.CommitID = c.CommitID`, and the other commands would use the updated value (it is memoized so it's not overwritten each time). But that means changing a global var. On balance, I think your way is best given the current state of things.

Issue #149, opened 2015-05-21
Windows support


pauljz:
It'd be great to have Windows support in srclib to support cross-platform editors.

Go itself supports Windows just fine, so most of the fixes that need to take place for this are oriented around taking advantage of Go's standard library. Most of the issue are simple things:

* Using `filepath.SplitList(paths)` instead of `strings.Split(paths, ":")`
* Using `os.PathSeparator` and `os.PathListSeparator ` instead of `":"` and `"/"`
* Using `os.TempDir()` to get temp paths instead of hardcoded `/tmp/` paths.

There are some more complex issues:

* srclib relies on `sh -c` for executing certain commands.
* srclib uses symlinks heavily, which while they work on Windows aren't terribly idiomatic. And by default cannot be created by a non-elevated process. (e.g. equivalent of having to `sudo` to symlink, ugh). We may need to provide a fallback that uses directory copies.
* Some of srclibs dependencies use different formats for command line flags `/methods:program` instead of `--methods=program`.
* And then of course there's Docker, which Windows at least now has some nascent support for. It's not clear to me how much will need to change to support this.

I've started on some fixes for a lot of these issues and will put a PR together soon once I'm sure the tests pass. I'll probably need a lot of help from folks that are more experienced with the srclib ecosystem, since this is all new to me.

shurcooL:
Hey @pauljz,

Windows support sounds great! The easy parts like using `filepath.SplitList` should definitely be done regardless, since it's more correct. I've recently made a PR https://github.com/sourcegraph/srclib-go/pull/35 to fix some of those. There are likely more things like that remaining.

I don't have a Windows machine to test on, but I can try to assist in other ways. I'll keep an eye on this and help review your future PRs.

pauljz:
This round of changes should be ready to review now, @shurcooL. I took out the changes to toolchain_cmd that were breaking the build, changed all refs to `os.PathSeparator` to `filepath.Separator` for consistency, and changed erroneous `path.Join`s to `filepath.Join`s.

Travis build passes, all tests pass locally now on OS X for me. Windows tests... well, still a long way to go there, but it's better than it was.

PR #150, opened 2015-05-22
Windows fixes

pauljz:
First round of Windows fixes. So far this is mostly simple fixes like dealing with path separators and path list separators properly.

There's some iffy stuff in the toolchain command code (which I also couldn't get working just yet due to symlink issues), but most of the rest of it should be pretty safe. I'll call that out in the comments on those lines.

The tests all pass locally for me on *nix, but that doesn't mean it's all happily working. Warrants some close scrutiny just due to how many things this touches.

Part of #149

### DIFF
--- a/env.go
+++ b/env.go
@@ -4,7 +4,6 @@ import (
 	"log"
 	"os"
 	"path/filepath"
-	"strings"
 
 	"sourcegraph.com/sourcegraph/srclib/util"
 )
@@ -35,7 +34,7 @@ func init() {
 	}
 
 	if CacheDir == "" {
-		dirs := strings.SplitN(Path, ":", 2)
+		dirs := filepath.SplitList(Path)
 		CacheDir = filepath.Join(dirs[0], ".cache")
 	}
 }
--- a/graph/repo.go
+++ b/graph/repo.go
@@ -4,7 +4,7 @@ import (
 	"errors"
 	"fmt"
 	"net/url"
-	"path/filepath"
+	"path"
 	"strings"
 )
 
@@ -35,10 +35,10 @@ func TryMakeURI(cloneURL string) (string, error) {
 		return "", fmt.Errorf("MakeURI(%q): %s", cloneURL, err)
 	}
 
-	path := strings.TrimSuffix(url.Path, ".git")
-	path = filepath.Clean(path)
-	path = strings.TrimSuffix(path, "/")
-	return strings.ToLower(url.Host) + path, nil
+	uri := strings.TrimSuffix(url.Path, ".git")
+	uri = path.Clean(uri)
+	uri = strings.TrimSuffix(uri, "/")
+	return strings.ToLower(url.Host) + uri, nil
 }
 
 // URIEqual returns true if a and b are equal, based on a case insensitive
--- a/plan/data.go
+++ b/plan/data.go
@@ -1,7 +1,6 @@
 package plan
 
 import (
-	"fmt"
 	"path/filepath"
 
 	"sourcegraph.com/sourcegraph/srclib/buildstore"
@@ -13,5 +12,5 @@ func RepositoryCommitDataFilename(emptyData interface{}) string {
 }
 
 func SourceUnitDataFilename(emptyData interface{}, u *unit.SourceUnit) string {
-	return filepath.Clean(fmt.Sprintf("%s/%s.%s", u.Name, u.Type, buildstore.DataTypeSuffix(emptyData)))
+	return filepath.Join(u.Name, u.Type+"."+buildstore.DataTypeSuffix(emptyData))
 }
--- a/src/cli.go
+++ b/src/cli.go
@@ -5,6 +5,7 @@ import (
 	"net/http"
 	"net/url"
 	"os"
+	"path/filepath"
 
 	"strconv"
 
@@ -59,7 +60,8 @@ func newAPIClient(ua *userEndpointAuth, cache bool) *sourcegraph.Client {
 
 	var transport http.RoundTripper
 	if cache {
-		transport = http.RoundTripper(httpcache.NewTransport(diskcache.New("/tmp/srclib-cache")))
+		tempDir := filepath.Join(os.TempDir(), "srclib-cache")
+		transport = http.RoundTripper(httpcache.NewTransport(diskcache.New(tempDir)))
 	} else {
 		transport = http.DefaultTransport
 	}
--- a/src/lint_cmd.go
+++ b/src/lint_cmd.go
@@ -102,10 +102,10 @@ func (c *LintCmd) Execute(args []string) error {
 						// Infer source unit name from file path (the
 						// path components after .srclib-cache until
 						// the basename).
-						pcs := strings.Split(absPath, string(os.PathSeparator))
+						pcs := strings.Split(absPath, string(filepath.Separator))
 						for i, pc := range pcs {
 							if pc == buildstore.BuildDataDirName && len(pcs) > i+2 {
-								unitName = filepath.Clean(strings.Join(pcs[i+2:len(pcs)-1], "/"))
+								unitName = filepath.Join(pcs[i+2 : len(pcs)-1]...)
 								break
 							}
 						}
@@ -114,7 +114,7 @@ func (c *LintCmd) Execute(args []string) error {
 					var commitID string
 					if !c.NoCheckFiles {
 						// Infer commit ID from file path (the path component after .srclib-cache).
-						pcs := strings.Split(absPath, string(os.PathSeparator))
+						pcs := strings.Split(absPath, string(filepath.Separator))
 						for i, pc := range pcs {
 							if pc == buildstore.BuildDataDirName && len(pcs) > i+1 {
 								commitID = pcs[i+1]
--- a/src/toolchain_cmd.go
+++ b/src/toolchain_cmd.go
@@ -414,7 +414,7 @@ func installGoToolchain() error {
 		return skippedToolchain{toolchain, "no GOPATH set (assuming Go is not installed and you don't want the Go toolchain)"}
 	}
 
-	srclibpathDir := filepath.Join(strings.Split(srclib.Path, ":")[0], toolchain) // toolchain dir under SRCLIBPATH
+	srclibpathDir := filepath.Join(filepath.SplitList(srclib.Path)[0], toolchain) // toolchain dir under SRCLIBPATH
 
 	if err := os.MkdirAll(filepath.Dir(srclibpathDir), 0700); err != nil {
 		return err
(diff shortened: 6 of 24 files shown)


src/toolchain_cmd.go
```diff
@@ -414,9 +414,9 @@ func installGoToolchain() error {
 		return skippedToolchain{toolchain, "no GOPATH set (assuming Go is not installed and you don't want the Go toolchain)"}
 	}
 
-	srclibpathDir := filepath.Join(strings.Split(srclib.Path, ":")[0], toolchain) // toolchain dir under SRCLIBPATH
+	srclibpathDir := filepath.Join(filepath.SplitList(srclib.Path)[0], toolchain) // toolchain dir under SRCLIBPATH
 
-	if err := os.MkdirAll(filepath.Dir(srclibpathDir), 0700); err != nil {
```
pauljz: I'm not sure why this `filepath.Dir` was here. On Windows it actually removes the toolchain dir and makes it so the code doesn't make any sense. We immediately try to symlink a folder that we didn't create.

I'm thinking on *nix there's a trailing slash and this just ends up trimming the trailing slash?

This happens on line 551 too.

src/toolchain_cmd.go
```diff
@@ -542,17 +542,18 @@ func symlinkToGopath(toolchain string) (skip string, err error) {
...
 
-	srcDir := filepath.Join(strings.Split(gopath, ":")[0], "src")
+	srcDir := filepath.Join(filepath.SplitList(gopath)[0], "src")
 	gopathDir := filepath.Join(srcDir, toolchain)
-	srclibpathDir := filepath.Join(strings.Split(srclib.Path, ":")[0], toolchain)
+	srclibpathDir := filepath.Join(filepath.SplitList(srclib.Path)[0], toolchain)
 
 	if fi, err := os.Lstat(gopathDir); os.IsNotExist(err) {
 		log.Printf("mkdir -p %s", filepath.Dir(gopathDir))
-		if err := os.MkdirAll(filepath.Dir(gopathDir), 0700); err != nil {
```
pauljz: Psst, over here.

toolchain/find.go
```diff
@@ -133,11 +134,16 @@ func newInfo(toolchainPath, dir, configFile string) (*Info, error) {
...
+	if runtime.GOOS == "windows" {
+		prog += ".exe"
+	}
+
 	if fi, err := os.Stat(filepath.Join(dir, prog)); os.IsNotExist(err) {
 		prog = ""
 	} else if err != nil {
 		return nil, err
-	} else if !(fi.Mode().Perm()&0111 > 0) {
+	} else if runtime.GOOS != "windows" && !(fi.Mode().Perm()&0111 > 0) {
```
pauljz: Since Windows doesn't have the same model of permissions as *nix this freaks out. We can generally assume on Windows that a .exe is going to be executable though.

src/cli.go
```diff
@@ -59,7 +60,8 @@ func newAPIClient(ua *userEndpointAuth, cache bool) *sourcegraph.Client {
 
 	var transport http.RoundTripper
 	if cache {
-		transport = http.RoundTripper(httpcache.NewTransport(diskcache.New("/tmp/srclib-cache")))
+		tempdir := path.Join(os.TempDir(), "srclib-cache")
```
pauljz: Should this use `ioutil.TempDir` instead? Wasn't sure if this needs to persist between requests or some such in which case `ioutil.TempDir`'s ability to pick a different name if the one you ask for is taken would be a bad thing.

graph/repo.go
```diff
@@ -35,10 +35,10 @@ func TryMakeURI(cloneURL string) (string, error) {
 		return "", fmt.Errorf("MakeURI(%q): %s", cloneURL, err)
 	}
 
-	path := strings.TrimSuffix(url.Path, ".git")
-	path = filepath.Clean(path)
```
pauljz: This looks weird but since these are URLs not filepaths, using `filepath.Clean` was actually replacing `/`s in the URL with `\`s.

src/cli.go
shurcooL: I'm unsure if the `srclib-cache` is actually reused, but we can ignore that for now and just preserve the behavior as before (while making it cross platform).

So using `os.TempDir()` is fine.

But you should be using `filepath.Join` here, not `path.Join`, since you're joining a system filepath.

graph/repo.go
shurcooL: Yep, this is a good change. URLs are always forward-slash separated paths, hence `path` package should be used, not `filepath`.

src/cli.go
```diff
@@ -59,7 +60,8 @@ func newAPIClient(ua *userEndpointAuth, cache bool) *sourcegraph.Client {
 
 	var transport http.RoundTripper
 	if cache {
-		transport = http.RoundTripper(httpcache.NewTransport(diskcache.New("/tmp/srclib-cache")))
+		tempdir := filepath.Join(os.TempDir(), "srclib-cache")
```
pauljz: Oops, doing a force push deleted the conversation on this line. This was erroneously `path.Join` instead of `filepath.Join` in an older version.

shurcooL: No problem. Please ping me once you've done with the comments I posted here and in https://github.com/sourcegraph/srclib-go/pull/38. I saw there were a few instances of `path.Join` being used where `filepath.Join` should be used instead.

Then I'll review all of your changes again. Thanks!

plan/data.go
```diff
@@ -13,5 +13,5 @@ func RepositoryCommitDataFilename(emptyData interface{}) string {
 }
 
 func SourceUnitDataFilename(emptyData interface{}, u *unit.SourceUnit) string {
-	return filepath.Clean(fmt.Sprintf("%s/%s.%s", u.Name, u.Type, buildstore.DataTypeSuffix(emptyData)))
+	return filepath.Clean(fmt.Sprintf("%s%s%s.%s", u.Name, string(filepath.Separator), u.Type, buildstore.DataTypeSuffix(emptyData)))
```
shurcooL: At this point, the value of using `fmt.Sprintf` is getting pretty low. If you want, you could rewrite this using string concatenation instead.

Either way is fine though.

Or maybe it could be `filepath.Join(u.Name, u.Type+"."+buildstore.DataTypeSuffix(emptyData))`.

src/lint_cmd.go
```diff
@@ -102,10 +102,10 @@ func (c *LintCmd) Execute(args []string) error {
 						// Infer source unit name from file path (the
 						// path components after .srclib-cache until
 						// the basename).
-						pcs := strings.Split(absPath, string(os.PathSeparator))
+						pcs := strings.Split(absPath, string(filepath.Separator))
 						for i, pc := range pcs {
 							if pc == buildstore.BuildDataDirName && len(pcs) > i+2 {
-								unitName = filepath.Clean(strings.Join(pcs[i+2:len(pcs)-1], "/"))
+								unitName = filepath.Clean(strings.Join(pcs[i+2:len(pcs)-1], string(filepath.Separator)))
```
shurcooL: Perhaps use `filepath.Join(x...)` instead of `strings.Join(x, string(filepath.Separator))` here?

toolchain/add.go
```diff
@@ -31,7 +30,7 @@ func Add(dir, toolchainPath string, opt *AddOpt) error {
 		return err
 	}
 
-	srclibpathEntry := strings.SplitN(srclib.Path, ":", 2)[0]
+	srclibpathEntry := filepath.SplitList(srclib.Path)[0]
```
shurcooL: If you're going to do this here, it'd be more consistent to also do it above (e.g., see https://github.com/sourcegraph/srclib/pull/150/files#diff-3e59f888b6806ff36d242aaa240cb59aL33).

It's also worth noting there is a slight deviation in behavior between the two funcs (highlighted in bold):

<pre>
// SplitList splits a list of paths joined by the OS-specific ListSeparator,
// usually found in PATH or GOPATH environment variables.
<strong>// Unlike strings.Split, SplitList returns an empty slice when passed an empty string.</strong>
func SplitList(path string) []string {
	return splitList(path)
}
</pre>

I think it's likely fine, but I wanted to point it out.

shurcooL:
Overall, looks good. I've left 3 comments above.

pauljz:
Fixed those up. Does the error `errors.New("SRCLIBPATH not set")` look okay? Should be clearer than blowing up trying to access an element of an empty slice. :)

If everything looks good, let me know and I can squash this before you merge.

env.go
```diff
@@ -35,7 +35,7 @@ func init() {
 	}
 
 	if CacheDir == "" {
-		dirs := strings.SplitN(Path, ":", 2)
+		dirs := strings.SplitN(Path, string(filepath.ListSeparator), 2)
```
shurcooL: For consistency with above, this should become:

```Go
dirs := filepath.SplitList(Path)
```

src/cli.go
shurcooL: Minor, but shouldn't this local variable be named `tempDir` to follow idiomatic Go style?

src/toolchain_cmd.go
```diff
@@ -414,7 +414,7 @@ func installGoToolchain() error {
 		return skippedToolchain{toolchain, "no GOPATH set (assuming Go is not installed and you don't want the Go toolchain)"}
 	}
 
-	srclibpathDir := filepath.Join(strings.Split(srclib.Path, ":")[0], toolchain) // toolchain dir under SRCLIBPATH
+	srclibpathDir := filepath.Join(filepath.SplitList(srclib.Path)[0], toolchain) // toolchain dir under SRCLIBPATH
```
shurcooL: Do these also need a `if srclib.Path == "" { return errors.New("srclib.Path is empty") }` guard?

shurcooL:
Hey @pauljz,

I've thoroughly reviewed the latest version of this PR (commit f7e36abfcdeb8b0a5e55396bb65b06120e0b8417), and overall it looks good. I've left 3 comments above. Two of them are easy to address. Please avoid squashing for now (so it's easier for me to review the upcoming changes).

The third comment, about the `srclib.Path == ""` guard is trickier.

```diff
-strings.SplitN(srclib.Path, ":", 2)[0]
+filepath.SplitList(srclib.Path)[0]
```

Looking through the code, there are some places that do that check to avoid a possible panic, and there are still 5-8 other places that do not.

I think we should be consistent about that, but I'm not sure if adding 5-8 more guards is the way to go. Maybe it is. Do you have thoughts on this?

Thanks for your patience as we work out the remaining issues!

shurcooL:
> Looking through the code, there are some instances of the above code that do that check to avoid a possible panic, and there are still 5-8 other places that do not.
> 
> I think we should be consistent about that, but I'm not sure if adding 5-8 more guards is the way to go. Maybe it is. Do you have thoughts on this?

I'm leaning in the direction of just documenting that `srclib.Path` should not be empty as a precondition, and not add any guards. If `srclib.Path` is ever empty, even with current code, unexpected things will happen (but no panic). I think it should never be empty (there is code ensuring that), and if it happens to be, then a panic is an acceptable outcome, since something must be seriously wrong for that to ever happen.

pauljz:
Yeah, given the checks that already happen in `env.go` it seems awfully redundant to have guards scattered throughout the code. I took out the other guards in this latest commit.

The other consistency things are all updated too.

shurcooL:
Thanks, that looks great!

LGTM. This is good to merge now. There shouldn't be any behavior changes on OS X and Linux, but Windows (and any other OSes supported by Go) behavior should be better.

pauljz:
Okay, squashed that last commit too.

shurcooL:
Thank you @pauljz!

PR #154, opened 2015-05-28
Make `src make` more informative

keegancsmith:
* --verbose flag for make and makefile subcommands
* Extra logging around target creation in verbose mode
* Final output line with colourized info indicating build success or failure


shurcooL:
These changes are excellent! :+1:

PR #155, opened 2015-06-04
Rename binary from src to srclib, remove Sourcegraph API client subcommands

sqs:


cli/doall_cmd_test.go
```diff
@@ -45,7 +45,7 @@ func TestDoAll_cached_sample(t *testing.T) {
 	if o, err := gitCmd.CombinedOutput(); err != nil {
 		t.Fatal(string(o), err)
 	}
-	srcCmd = exec.Command("src", "do-all")
+	srcCmd = exec.Command("srclib", "do-all")
```
gbbr: I think this was an omission from a previous changeset, but I think this should be `srclib.CommandName`.

cli/doall_cmd_test.go
```diff
@@ -56,7 +56,7 @@ func TestDoAll_cached_sample(t *testing.T) {
...
```
gbbr: Same here

cli/makefile_cmd.go
```diff
@@ -9,8 +9,8 @@ import (
 
 func init() {
 	_, err := CLI.AddCommand("makefile",
-		"prints the Makefile that `src make` executes",
-		"The makefile command prints the Makefile that `src make` will execute.",
+		"prints the Makefile that `srclib make` executes",
+		"The makefile command prints the Makefile that `srclib make` will execute.",
```
gbbr: Also here, but feel free to leave as is

cli/selfupdate_cmd.go
```diff
@@ -30,8 +30,8 @@ func (c *SelfUpdateCmd) Execute(_ []string) error {
 		ApiURL:         url,
 		BinURL:         url,
 		DiffURL:        url,
-		Dir:            "src",
-		CmdName:        "src",
+		Dir:            "srclib",
+		CmdName:        "srclib",
```
gbbr: And here

gbbr:
This LGTM, I especially like the removal of the API calls and having srclib decoupled of that.

MaikuMori:
This is going to affect all the editor plugins.

I see you also removed `src q`, is it gone completely or will be somewhere in some form?

sqs:
@MaikuMori I forgot to mark this as WIP. You are completely right.

The editor plugins currently rely on a combination of local srclib data and remote Sourcegraph data. That's not ideal for a lot of reaosns. I think the best temporary solution is to make them rely on only local srclib data (so no global usage examples, for example). This means basically adding back the `api` subcommands, except with slightly modified types so they don't depend on go-sourcegraph (we should ensure the JSON output is the same).

As for `src q`, that can be added back. I removed it to quickly prototype getting this branch to work. Do you currently use `src q`?

Another breaking change is the rename from `src` to `srclib`, if that occurs. The "nodb" branch, when finally merged, will be a significant version bump and we will be sure to update all editor plugins and give appropriate notice.

MaikuMori:
The global examples feature is probably the most interesting/wanted for the editor plugins followed by jump to definition (this can be local since you typically have library files on the pc).

I don't use `src q` because it has never worked for me. Something about missing formatter. If it would work then it could be handy when debugging plugins/toolchains. Since it never worked it's not catastrophic that it's gone.

What is `nodb` branch about?

sqs:
> The global examples feature is probably the most interesting/wanted for the editor plugins followed by jump to definition (this can be local since you typically have library files on the pc).

Yes, you are right. Let me think about how to deal with this then... We could have the editor plugins call a separate binary (in the interest of keeping srclib self contained).

> What is nodb branch about?

It's related to an internal refactor to make Sourcegraph not rely on PostgreSQL (thus "no DB").

MaikuMori:
That sounds Okey.

sqs:
Added back `src q` and `src api ...` commands (with no dependency on Sourcegraph.com). I am going to merge this into nodb and we will address editor plugin compat issues (binary renaming from src to srclib) when we merge nodb into master. Thanks for the comments.

PR #158, opened 2015-06-05
graph: Handle file paths passed to `TryMakeURI`

keegancsmith:
If a user clones a local repo with `git clone /path/to/repo /path/to/destrepo` the remote URL is a path `/path/to/destrepo`. We didn't error out from `TryMakeURI` on these inputs.

### DIFF
--- a/graph/repo.go
+++ b/graph/repo.go
@@ -35,7 +35,7 @@ func TryMakeURI(cloneURL string) (string, error) {
 		return "", fmt.Errorf("MakeURI(%q): %s", cloneURL, err)
 	} else if url.Path == "" || url.Path == "/" {
 		return "", fmt.Errorf("MakeURI(%q): missing path from URL", cloneURL)
-	} else if url.Host == "" && !strings.Contains(strings.Split(url.Path, "/")[0], ".") {
+	} else if url.Host == "" && (url.Path[0] == '/' || !strings.Contains(strings.Trim(url.Path, "/"), "/")) {
 		// We ensure our Path doesn't look like the output of TryMakeURI
 		// so that the output of this function is a fixed point.
 		// ie TryMakeURI("github.com/user/repo") == ("github.com/user/repo", nil),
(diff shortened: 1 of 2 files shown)


graph/repo.go
```diff
@@ -33,6 +33,14 @@ func TryMakeURI(cloneURL string) (string, error) {
...
 	if err != nil {
 		return "", fmt.Errorf("MakeURI(%q): %s", cloneURL, err)
+	} else if url.Path == "" || url.Path == "/" {
+		return "", fmt.Errorf("MakeURI(%q): missing path from URL", cloneURL)
+	} else if url.Host == "" && !strings.Contains(strings.Split(url.Path, "/")[0], ".") {
+		// We ensure our Path doesn't look like the output of TryMakeURI
+		// so that the output of this function is a fixed point.
+		// ie TryMakeURI("github.com/user/repo") == ("github.com/user/repo", nil),
+		// not an error.
+		return "", fmt.Errorf("MakeURI(%q): missing host from URL", cloneURL)
```
sqs: These errors may be displayed to the user. If the user is not familiar with the code (or doesn't have access to it--e.g., they only have a binary), then the `MakeURI(%q):` prefix will be confusing and not-helpful. I would suggest just using error messages like `missing host from URL (%q)` (or `determining URI from repo clone URL failed: missing host from URL (%q)`) here and just passing through the error on line 35.

graph/repo.go
```diff
@@ -33,6 +33,14 @@ func TryMakeURI(cloneURL string) (string, error) {
 	url, err := url.Parse(cloneURL)
 	if err != nil {
 		return "", fmt.Errorf("MakeURI(%q): %s", cloneURL, err)
+	} else if url.Path == "" || url.Path == "/" {
+		return "", fmt.Errorf("MakeURI(%q): missing path from URL", cloneURL)
+	} else if url.Host == "" && !strings.Contains(strings.Split(url.Path, "/")[0], ".") {
```
sqs: This assumes that the hostname has a dot, which is not true in the general case (e.g., for localhost, or internal server hostnames). While it's unlikely that someone would want a repo URI like `localhost/foo/bar`, it's reasonable that they'd have `http://gitrepos` be an intranet host and thus would have `gitrepos/foo/bar` be a repo URI. 

I think it's fine to adopt the following lesser constraint: the host + "/" + path must not be empty. If it is empty, then return an error.

keegancsmith:
@sqs Addressed both your comments and added some extra unit tests.

Issue #171, opened 2015-06-30
src api units and src api deps shouldn't remake


xizhao:
These are accessors for build data info, they should not be remaking the repo.

Currently because I am omitting the graph op, remake fails and it clears the srclib cache
https://github.com/sourcegraph/srclib/blob/7e6ce0746ea6043833d3a229085dffc16c2c8c3c/src/api_cmds.go#L189

I am migrating my stuff to the `build-data` api cmds

PR #172, opened 2015-06-30
#171 Disable ensureBuild for `src api units` and `src api deps`

xizhao:


src/api_cmds.go
```diff
@@ -139,7 +139,7 @@ type commandContext struct {
 // creates commandContext, changes the process' working directory to
 // file's directory and ensures that a build has been made. It is
 // meant to be used by user-facing commands.
-func prepareCommandContext(file string) (commandContext, error) {
+func prepareCommandContext(file string, ensure_build bool) (commandContext, error) {
```
shurcooL: Go style uses mixed caps casing for identifiers, see [here](https://github.com/golang/go/wiki/CodeReviewComments#mixed-caps).

Also, the documentation should be updated accordingly, since this is no longer always true:

> and ensures that a build has been made.

Issue #103, opened 2015-01-18
Src pull recent or all remote build files


xizhao:
Pull the most recent remote build file in order to take advantage of incremental builds from last known commit

It would be nice to be able to pull all remote files as well in order to have a better commit log.

xizhao:
Bump.  Ideas:

`src ls-remote` - list all commits from most recent to least
`src fetch --recent` - fetch the most recent commit

xizhao:
`src fetch` should fetch the most recent commit by default in order to leverage incremental build features, or perhaps do-all

PR #173, opened 2015-06-30
#103 add --latest flag to fetch latest build data

xizhao:
This should query for the last successful build and download the latest, rather than failing if the current is not found.

### DIFF
--- a/src/build_data_cmds.go
+++ b/src/build_data_cmds.go
@@ -397,6 +397,7 @@ type BuildDataFetchCmd struct {
 	buildDataSingleCommitCommonOpts
 
 	DryRun bool `short:"n" long:"dry-run" description:"don't do anything, just show what would be done"`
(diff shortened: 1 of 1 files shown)


src/build_data_cmds.go
```diff
@@ -397,6 +397,7 @@ type BuildDataFetchCmd struct {
 	buildDataSingleCommitCommonOpts
 
 	DryRun bool `short:"n" long:"dry-run" description:"don't do anything, just show what would be done"`
+	Latest bool `short:"l" long:"latest" description: "fetch the latest successful build"`
```
shurcooL: I don't think you're allowed to put a space between `description:` and `"fetch the latest successful build"`, that will probably not parse correctly.

[Go documentation](https://godoc.org/reflect#StructTag) states:

> By convention, tag strings are a concatenation of optionally space-separated key:"value" pairs.

xizhao:
Sorry, I pushed the wrong commit.  This one doesn't build.

PR #174, opened 2015-06-30
toolchain: add support for prebuilt toolchain bundles

sqs:
Toolchain bundles are .tar.gz archives of precompiled srclib toolchains.
Binary distribution of toolchains allows people to install and use toolchains
without needing to compile them (which often requires many
difficult-to-install dependencies).

Toolchains define how their bundles are created and what bundle variants (for
cross-compilation) should be built. Then users can install the bundles by
downloading the .tar.gz archive and running
'srclib toolchain unbundle <FILE>'.

**Dependent PRs**
* https://github.com/sourcegraph/srclib-go/pull/40
* https://github.com/sourcegraph/srclib-java/pull/30

### DIFF
--- a/cli/toolchain_cmd.go
+++ b/cli/toolchain_cmd.go
@@ -3,10 +3,16 @@ package cli
 import (
 	"errors"
 	"fmt"
+	"io"
+	"io/ioutil"
 	"log"
+	"net/http"
+	"net/url"
 	"os"
 	"os/exec"
+	"path"
 	"path/filepath"
+	"time"
 
 	"strings"
 	"sync"
@@ -59,6 +65,24 @@ func init() {
 		"download a toolchain",
 		"Download a toolchain's repository to the SRCLIBPATH.",
 		&toolchainGetCmd,
+	)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	_, err = c.AddCommand("bundle",
+		"bundle a toolchain",
+		"The bundle subcommand builds and archives toolchain bundles (.tar.gz files, one per toolchain variant). Bundles contain prebuilt toolchains and allow people to use srclib toolchains without needing to compile them on their own system.",
+		&toolchainBundleCmd,
+	)
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	_, err = c.AddCommand("unbundle",
+		"unbundle a toolchain",
+		"The unbundle subcommand unarchives a toolchain bundle (previously created with the 'bundle' subcommand). It allows people to download and use prebuilt toolchains without needing to compile them on their system.",
+		&toolchainUnbundleCmd,
 	)
 	if err != nil {
 		log.Fatal(err)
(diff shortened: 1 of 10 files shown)


shurcooL:
What do you think about making this the default, and installing toolchains from source a secondary, more advance use case?

The current API seems rather confusing and not well organized from a user's perspective IMO.

- `srclib toolchain bundle` - As I understand, this is a developer-oriented command that creates a new bundle. Yet it uses up a shorter and nicer word "bundle".

- `srclib toolchain unbundle` - As I understand, this is a user-oriented command for unpacking/installing an existing bundle. This should have the nicest user-facing API, yet it uses word "unbundle", and I'm unsure how clear it is.

This is a high level thought, but is there something stopping this from being the default, easier way that all users install toolchains? If they wish to build from source, that IMO should be a secondary way to do it (since it takes more time, dependencies and expertise). If so, I'd suggest we reuse/replace the existing API to install packages:

`src toolchain install foo` - Currently installs foo toolchain from source. Why not make it download and install from bundle, unless you opt to install from source (perhaps as a flag, or through another command).

> Then users can install the bundles by
downloading the .tar.gz archive and running
'srclib toolchain unbundle '.

So the current behavior requires users to download a bundle (from a website, I presume) first, and then run unbundle, right? To me, this screams of a simpler command that does both tasks (downloads bundle and installs it).

These are high level thoughts, feel free to consider them outside of this PR. But it just seems if we're going for ease of install, convenience and speed, this approach deserves much nicer user facing API, doesn't it?

sqs:
I agree. This should be the default.

sqs:
OK, for now I have made this binary bundle installation the default if you run `srclib toolchain install java`. 

@shurcooL, can you please review this? Thanks!

cli/toolchain_cmd.go
```diff
@@ -522,25 +616,75 @@ func installPythonToolchain() error {
...
+			}
+		}
+	}()
 
-	srclibpathDir := filepath.Join(strings.Split(srclib.Path, ":")[0], toolchain) // toolchain dir under SRCLIBPATH
-	log.Println("Downloading or updating Java toolchain in", srclibpathDir)
-	if err := execSrcCmd("toolchain", "get", "-u", toolchain); err != nil {
+	log.Printf("Downloading %s (%.1f MB)", bundleURL, float64(resp.ContentLength)/1024/1024)
+	_, err = io.Copy(f, resp.Body)
+	done <- struct{}{}
```
shurcooL: Pretty minor, I would move the `log.Printf("Finished downloading")` line from the goroutine above to here. Otherwise, it may end up printing a little later (since this is a one-way synchronization) and looking odd:

```
Downloading https://srclib-support.s3-us-west-2.amazonaws.com/...tar.gz (356.4 MB)
Download 99.6% complete (354.9 MB / 356.4 MB)
Unarchiving Java toolchain bundle at /var/folders/_w/3kl_g4rn5sv1nyjblk7tf4h00000gn/T/srclib-toolchain-bundle402629058/srclib-java__bundle__jdk-1.8.tar.gz
Finished downloading
```

shurcooL:
The new (user-facing) interface is great, very easy. Aside from my minor comment regarding out-of-place "Finished downloading" message, I have another comment:

It prints a lot of relevant information, but one missing piece seems the "Installed into <dir>" string:

```
$ srclib toolchain install java
Java (sourcegraph.com/sourcegraph/srclib-java) ================================
Installing Java toolchain from https://srclib-support.s3-us-west-2.amazonaws.com/srclib-java/srclib-java__bundle__jdk-1.8.tar.gz
Downloading https://srclib-support.s3-us-west-2.amazonaws.com/srclib-java/srclib-java__bundle__jdk-1.8.tar.gz (356.4 MB)
Download 99.6% complete (354.9 MB / 356.4 MB)
Unarchiving Java toolchain bundle at /var/folders/_w/3kl_g4rn5sv1nyjblk7tf4h00000gn/T/srclib-toolchain-bundle402629058/srclib-java__bundle__jdk-1.8.tar.gz
Finished downloading
OK! Installed/upgraded Java (sourcegraph.com/sourcegraph/srclib-java) toolchain
================================================================================
```

I would consider adding that since that would feed more natural.

Aside from that, it's absolutely great.

On the code review side, it's a large PR, but I've tried to read it all, and didn't spot any issues (aside what's already mentioned). Maybe consider using the [`go-humanize`](https://godoc.org/github.com/dustin/go-humanize) package instead of doing ad-hoc filesize calculations. I haven't confirmed the installed Java toolchain works, but it looks like it should.

LGTM.

PR #177, opened 2015-07-03
Make go generate directives work on more setups.

shurcooL:
Add support for running go generate successfully on OS X.

Add support for running go generate with multiple GOPATH workspaces.

Use `(gogoproto.customtype) = "encoding/json.RawMessage"` instead of doing it via custom sed string substitution.


neelance:
LGTM

shurcooL:
Let me verify this doesn't break a project that vendors this, and I'll merge.

PR #178, opened 2015-07-03
moved DefFormatStrings to srclib/graph

neelance:


graph/def.proto
```diff
@@ -116,3 +116,23 @@ message DefDoc {
     // Data is the actual documentation text.
     string data = 2 [(gogoproto.jsontag) = "Data"];
 };
+
+// DefFormatStrings contains the various def format strings.
```
sqs: Mention something like: for more information, see the godoc in the srclib/graph package.

sqs:
lgtm

shurcooL:
LGTM.

PR #181, opened 2015-07-23
Parallise over units for unitStore ref filters

keegancsmith:


store/unit_store.go
```diff
@@ -89,31 +89,50 @@ func (s unitStores) Refs(f ...RefFilter) ([]*graph.Ref, error) {
...
+		par.Do(func() error {
+			// Copy so we can do concurrent modification
+			fCopy := filtersForUnit(u, f).([]RefFilter)
+			setImpliedUnit(fCopy, u)
+
+			refs, err := us.Refs(fCopy...)
+
+			// Use a lock since increment isn't atomic
+			allRefsMu.Lock()
+			c_unitStores_Refs_last_numUnitsQueried++
```
beyang: non-atomic mutation -- looks like `c_unitStores_Refs_last_numUnitsQueried` is only used in tests, but this might cause that test to become flakey?

sqs: Those are statistics counters and the exact value is not important. They are also used only in tests, but even if they are used in main code, the exact value is not important. It was a specific design decision to not make those protected by a mutex (one that admittedly is not documented, sorry).

beyang: this block of code should also be moved to the immediately after `setImpliedUnit...` so that the error check for `refs, err := us.Refs(fCopy...)` can come right after.

store/unit_store.go
```diff
@@ -80,7 +81,10 @@ func (s unitStores) Defs(fs ...DefFilter) ([]*graph.Def, error) {
 	return allDefs, err
 }
 
-var c_unitStores_Refs_last_numUnitsQueried = 0
+var (
+	c_unitStores_Refs_last_numUnitsQueried = 0
+	c_mu                                   sync.Mutex
```
shurcooL: Style. It's common to put the mutex as a "hat" on top of the things it protects. So you should swap these two variables.

store/unit_store.go
```diff
@@ -89,31 +93,69 @@ func (s unitStores) Refs(f ...RefFilter) ([]*graph.Ref, error) {
 	}
 
 	c_unitStores_Refs_last_numUnitsQueried = 0
-	var allRefs []*graph.Ref
+	var (
+		allRefs   []*graph.Ref
+		allRefsMu sync.Mutex
```
shurcooL: Same as above.

store/unit_store.go
```diff
@@ -89,31 +93,69 @@ func (s unitStores) Refs(f ...RefFilter) ([]*graph.Ref, error) {
...
+	)
+	par := parallel.NewRun(storeFetchPar)
 	for u, us := range uss {
+		u, us := u, us
```
shurcooL: Put this line after the `if us == nil { continue }`, as that's better to keep as the topmost statement.

store/unit_store.go
```diff
@@ -89,31 +93,69 @@ func (s unitStores) Refs(f ...RefFilter) ([]*graph.Ref, error) {
...
-		for _, ref := range refs {
-			ref.UnitType = u.Type
-			ref.Unit = u.Name
-			if ref.DefUnitType == "" {
-				ref.DefUnitType = u.Type
+		par.Do(func() error {
+			// Copy so we can do concurrent modification in setImpliedUnit
+			fCopy := make([]RefFilter, len(f))
+			for i, filter := range f {
+				// HACK: consider adding a clone() method to filters
```
shurcooL: Consider changing the "HACK" into a "TODO" or implement it before merging.

store/unit_store.go
```diff
@@ -89,31 +93,69 @@ func (s unitStores) Refs(f ...RefFilter) ([]*graph.Ref, error) {
...
+				if ref.DefUnit == "" {
+					ref.DefUnit = u.Name
+				}
 			}
-		}
-		allRefs = append(allRefs, refs...)
+
+			allRefsMu.Lock()
+			allRefs = append(allRefs, refs...)
+			allRefsMu.Unlock()
```
shurcooL: While this works, is there a reason you didn't choose to go with using a channel sending `refs` across that, and collecting results at the end? Appending to a shared memory slice protected by a mutex seems quite unidiomatic.

shurcooL:
I've made a few style and code suggestions for possible improvements. Most are minor.

The only larger suggestion/comment I have is the way the parallelism is implemented, I think it would be nicer to do it more idiomatically by using a channel to collect the results rather than appending to a mutex-protected slice.

Otherwise I'm not spotting any other issues.

keegancsmith:
The reason it was done with locks is the method implements a similiar thing above, and it used locks. I'll clean up this PR to make it more idiomatic go and address your suggestions.

shurcooL:
That's completely understandable. The style suggestions can be a part of a separate followup PR; this PR can focus on functionality.

keegancsmith:
I'm not a go expert so I am likely doing something noobish and non-idiomatic, but using channels instead of a simple lock makes the code much more complex in my view. So I'll stick for now with the mutex protected version - https://gist.github.com/keegancsmith/4d6bec9f64ba2f665f4f

shurcooL:
I'm okay with that.

The diff looks good, and I don't see anything non-idiomatic about it, except the usage of `parallel.NewRun` package maybe. The Go standard library, for example, would not do that, it would just use the `sync.WaitGroup` primitive directly.

Personally, the channel version looks beautiful and simple to my eyes, even though it's longer amount of code. But that's because I'm used to this typo of Go code.

keegancsmith:
Ok, I'll try craft some code that pleases your eyes :)
/me is trying to level up his Go skills

keegancsmith:
Actually I'll follow up with more idiomatic Go at a later stage. The more I think about it the more I prefer the terseness/KISS of the current version.

shurcooL:
I think that's the best way to go. :) Leave it as is for now, but consider coming back to it at a later date. Refactoring is best done when it has a purpose and strong inner motivation.

sqs:
@keegancsmith, is this ready to merge?

PR #182, opened 2015-07-28
src make documentation correction

angusb:


docs/sources/api/make.md
```diff
@@ -50,7 +50,7 @@ necessary commands to analyze the project. To do so, it examines each source
 unit (using the build cache, if present) and generates rules for each of the
 predefined operations (currently depresolve and graph).
 
-You can view the Makefile by running `src make --print`.
+You can view the Makefile by running `src make --dry-run`.
```
beyang: Actually, this should be `src makefile` (which prints the Makefile, where as `--dry-run` prints a list of actions that will be executed, which might be empty if all the make targets exist already). Thanks for catching the typo, though!

angusb: Ah, got it!

PR #184, opened 2015-07-31
Remove unused temp-dir command

beyang:
@MaikuMori you added `TempDir` and the `temp-dir` subcommand a few months back, but it looks like this is as yet unused. Currently, all the toolchains just create their own temporary directories in the standard OS temp directory. This seems like a reasonable thing to do. Any objections to removing this?


beyang:
Ah, I see that there is a reference to this in `srclib-python`. This is problematic, because it introduces a Go dependency from a toolchain back to srclib-core. The toolchains should be standalone programs that have no dependency back to srclib core. Toolchains can be written in any language, and in this case, there is no way for a toolchain written in another language to access the temp directory. Any objection to changing the function `getTempPath()` in `srclib-python` to use `tempfile`?

MaikuMori:
None at all, I agree about keeping the dependencies simple.

PR #186, opened 2015-08-03
Merge nodb branch into master

beyang:
This will make srclib consistent with the currently deployed version of Sourcegraph


MaikuMori:
Would be nice to use some semantic versioning and add changelog (at least for `srclib`). Would make it easier to fix compatibility issues in toolchains and tools. Also just for sanity reasons to keep track with stuff.

PR #190, opened 2015-08-27
Merge nodb into master

beyang:
Merge nodb into master, after resolving conflicts.


shurcooL:
Question, what were the merge conflicts?

Now that we're merging `nodb` into master, we should fix it so that the build is green, right?

It looks like the Go, Python, Ruby and JavaScript toolchains are succeeding, but Java one is failing with:

```
Running  [src toolchain get -u sourcegraph.com/sourcegraph/srclib-java]
failed to install/upgrade Java (sourcegraph.com/sourcegraph/srclib-java) toolchain: command ["src" "toolchain" "get" "-u" "sourcegraph.com/sourcegraph/srclib-java"] failed: exec: "src": executable file not found in $PATH
The following toolchains were not installed:
Java (sourcegraph.com/sourcegraph/srclib-java)
FAILED: srclib toolchain install-std
make: *** [install-std-toolchains] Error 1
The command "make install-std-toolchains" exited with 2.
```

What will it take to get that resolved? It's not finding `src` in $PATH, but how are other toolchains succeeding despite that?

beyang:
Most of the conflicts were related to the Windows support that was added to master. Another conflict was that in `master`, the Java toolchain was downloaded as a bundle as opposed to installed from source (which is the cause of the test failure now, which I'll look into). All the other conflicts were straightforward.

PR #191, opened 2015-08-27
perf: Query limit filter in unitstore.Refs

keegancsmith:
If one of our filters is a `limiter`, then we can avoid querying additional
unit stores if our limit is reached. On a large synthetic repo (500 units) this
took query time from 15s to 6s. This was an FS backed store through sourcegraph
on my MBA. The limit used is 2. Most of the 6s time is dominated by
unmarshalling the index, rather than the queries.

### DIFF
--- a/store/unit_store.go
+++ b/store/unit_store.go
@@ -103,6 +103,9 @@ func (s unitStores) Refs(f ...RefFilter) ([]*graph.Ref, error) {
 		c_unitStores_Refs_last_numUnitsQueried++
 
 		par.Do(func() error {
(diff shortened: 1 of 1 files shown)


slimsag:
I'm only able to test on a repo of 200 units (not 500) but in my case it didn't help very much, taking the time from ~2.1s to ~1.9s. I presume this is the same thing I identified yesterday, which is that I am primarily IO bound. But it sounds like this helped significantly for you on your MBA which I presume has an SSD.

LGTM

PR #192, opened 2015-08-30
Added line numbers to refs & defs

gbbr:


shurcooL:
What about refs, defs that span multiple lines, how will this handle that situation? It might be sufficient, but I want to learn and confirm.

gbbr:
Most defs do. I am only interested in the first line (maybe I should call
this "StartLine"). This is so we can blame and get the author of a def.
FirstLine should suffice for that, although, end line should also be
easily obtainable. I think you are right, I will add that too, it
might make more sense.

gbbr:
I've added your suggestion. I like it. The information is already there so we should use it.  

A Def's signature (first line) could be authored by someone and 90% of its contents could be someone else. This way we could extract the person who has worked the most on it, as well as the person who authored the signature of the definition (the first line).

beyang:
It seems a little redundant to store both line numbers and character offsets to indicate definition location. I understand why it's useful in this case -- `git blame` takes a line number rather than a char offset. However, computing blame data is something we should do at the `srclib` level, instead of (I assume) computing on the fly. That data is immutable and there's no reason we can't associate authorship data with each definition at build time rather than at runtime. FYI, we used to do just that in `srclib`, but took that out to simplify the schema since we weren't using that data anymore. Now that there's a need for that, I think we should add back the definition authorship computation step. @sqs any other context here that I'm missing?

gbbr:
I like your suggestion mode. I'll close this. Let's not add `StartLine` and `EndLine` and indeed just add authorship information. I wasn't sure whether we're going this way (didn't wanna introduce a `git` dependency).

gbbr:
@beyang one thing to consider here is that there may be various authors on a definition. For example, the author of the function definition's signature may be different from the author of its body. Would it be worthwhile storing multiple authors?  

Just something to ponder on.

PR #198, opened 2015-09-09
cli: do not skip installation of toolchains

slimsag:
Instead, error out right away when they fail. If someone runs `srclib toolchain install-std`
then they expect every toolchain to be installed, not for some to be skipped
(the user may potentially miss the error among all the log output). Likewise if
the user wants to explicitly install a given toolchain (e.g. `srclib toolchain install go`)
they do not want it to skip installation due to an error.

For example, trying to install the Go toolchain without $GOPATH set now produces:

```
$ srclib toolchain install go
Go (sourcegraph.com/sourcegraph/srclib-go) ====================================
failed to install/upgrade Go (sourcegraph.com/sourcegraph/srclib-go) toolchain:
Refusing to install Go toolchain because there is no GOPATH environment variable
set.
Note: Please ensure that Go 1.4+ is installed and that $GOPATH is set.

FAILED: srclib toolchain install go
```

Instead of:

```
skipped sourcegraph.com/sourcegraph/srclib-go: no GOPATH set (assuming Go is not installed and you don't want the Go toolchain)
```

Which is rather confusing given that I said `srclib toolchain install go`, so why is it assuming that I don't want the Go toolchain? I explicitly asked for it.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 MAKEFLAGS+=--no-print-directory
 
-.PHONY: default install srclib release upload-release check-release install-std-toolchains test-std-toolchains
+.PHONY: default install srclib release upload-release check-release install-all-toolchains test-all-toolchains
 
 default: install
 
@@ -33,12 +33,12 @@ check-release:
 	echo; echo
 	@echo Released srclib $(V)
 
-install-std-toolchains:
+install-all-toolchains:
 	srclib toolchain install go python ruby javascript java
 
 toolchains ?= go javascript python ruby
 
-test-std-toolchains:
+test-all-toolchains:
 	@echo Checking that all standard toolchains are installed
 	for lang in $(toolchains); do echo $$lang; srclib toolchain list | grep srclib-$$lang; done
 
@@ -47,5 +47,5 @@ test-std-toolchains:
 	@echo Testing installation of standard toolchains in Docker if Docker is running
 	(docker info && make -C integration test) || echo Docker is not running...skipping integration tests.
 
-regen-std-toolchain-tests:
+regen-all-toolchain-tests:
 	for lang in $(toolchains); do echo $$lang; cd ~/.srclib/sourcegraph.com/sourcegraph/srclib-$$lang; srclib test --gen; done
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ check-release:
 	@echo Released srclib $(V)
 
 install-std-toolchains:
-	srclib toolchain install-std
+	srclib toolchain go python ruby javascript java
 
 toolchains ?= go javascript python ruby
 
--- a/Makefile
+++ b/Makefile
@@ -34,7 +34,7 @@ check-release:
 	@echo Released srclib $(V)
 
 install-std-toolchains:
-	srclib toolchain go python ruby javascript java
+	srclib toolchain install go python ruby javascript java
 
 toolchains ?= go javascript python ruby
 
--- a/cli/toolchain_cmd.go
+++ b/cli/toolchain_cmd.go
@@ -469,7 +469,7 @@ func installRubyToolchain() error {
 	srclibpathDir := filepath.Join(strings.Split(srclib.Path, ":")[0], toolchain) // toolchain dir under SRCLIBPATH
 
 	if _, err := exec.LookPath("ruby"); isExecErrNotFound(err) {
-		return errors.New("no `ruby` in PATH (assuming you don't have Ruby installed and you don't want the Ruby toolchain)")
+		return errors.New("no `ruby` in PATH (do you have Ruby installed properly?)")
 	}
 	if _, err := exec.LookPath("bundle"); isExecErrNotFound(err) {
 		return fmt.Errorf("found `ruby` in PATH but did not find `bundle` in PATH; Ruby toolchain requires bundler (run `gem install bundler` to install it)")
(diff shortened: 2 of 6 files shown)


beyang:
It's not clear to me that this is better behavior. The original thought is that some toolchains might fail to install for a variety of reasons (e.g., the user has some weird way of configuring Ruby). In that case, it would be better to continue to install as many toolchains as possible rather than aborting early and not installing anything else. Currently, it does return an error if any one toolchain fails to install; it just does so after trying to install all toolchains.

Was there a specific use case in mind (e.g., making CI go faster, or maybe the error is not totally visible to the user)? If making the error more user-visible was the motivation, there are other ways we can accomplish that without dropping the "best-effort" behavior, which I think is valuable, because toolchain installation will likely always not be 100% robust.

slimsag:
PTAL

I have pushed another commit to remove the `install-std` sub-command outright (see commit message for details why).

> It's not clear to me that this is better behavior. The original thought is that some toolchains might fail to install for a variety of reasons (e.g., the user has some weird way of configuring Ruby). In that case, it would be better to continue to install as many toolchains as possible rather than aborting early and not installing anything else. Currently, it does return an error if any one toolchain fails to install; it just does so after trying to install all toolchains.

I've updated the PR description to reflect the before/after of this change. The primary rational is behind the confusing error message (skipping installation of go toolchain even though I explicitly asked for it).

> Was there a specific use case in mind (e.g., making CI go faster, or maybe the error is not totally visible to the user)? If making the error more user-visible was the motivation, there are other ways we can accomplish that without dropping the "best-effort" behavior, which I think is valuable, because toolchain installation will likely always not be 100% robust.

1. If the user asks for `python` and `go` to be installed, they really want both. If the installation process fails due to one, having the other is not very useful and in general the user may miss the error message and end up with srclib working on only half their code.
2. Installation failures are not in general a situation we should 'continue trying anyway' on -- if there is a serious fault in the installation process then we want users to raise issues about it so that we can make the installation process both easy and robust.

These are just my understanding of the context, but I'll ask @rothfels or @sqs to chime in and provide more context around the motivation of this change here.

rothfels:
@beyang given we decided the notion of "standard toolchain" was arbitrary we didn't feel it was worth keeping that bundled installation around. Which leaves the only question "what do we do when we fail to install a (single) toolchain"? Which I think Stephen has answered quite well. :)

Makefile
```diff
@@ -34,7 +34,7 @@ check-release:
 	@echo Released srclib $(V)
 
 install-std-toolchains:
```
rothfels: I thought we were removing this entirely?

rothfels: I thought we were removing this...

slimsag: @rothfels this is just part of the Makefile, so it won't end up as part of the srclib command. There is also `test-std-toolchains`, which I suspect we need. I'll rename them to something more appropriate though.

cli/toolchain_cmd.go
```diff
@@ -426,60 +417,27 @@ func (c *ToolchainInstallCmd) Execute(args []string) error {
...
-
 func installGoToolchain() error {
 	const toolchain = "sourcegraph.com/sourcegraph/srclib-go"
 	gopath := os.Getenv("GOPATH")
 	if gopath == "" {
-		return skippedToolchain{toolchain, "no GOPATH set (assuming Go is not installed and you don't want the Go toolchain)"}
+		return errors.New(`
+Refusing to install Go toolchain because there is no GOPATH environment variable
+set.
+Note: Please ensure that Go 1.4+ is installed and that $GOPATH is set.`)
```
rothfels: Add link to https://golang.org/doc/install

cli/toolchain_cmd.go
```diff
@@ -513,7 +469,7 @@ func installRubyToolchain() error {
 	srclibpathDir := filepath.Join(strings.Split(srclib.Path, ":")[0], toolchain) // toolchain dir under SRCLIBPATH
 
 	if _, err := exec.LookPath("ruby"); isExecErrNotFound(err) {
-		return skippedToolchain{toolchain, "no `ruby` in PATH (assuming you don't have Ruby installed and you don't want the Ruby toolchain)"}
+		return errors.New("no `ruby` in PATH (assuming you don't have Ruby installed and you don't want the Ruby toolchain)")
```
rothfels: Isn't this error message going to cause the confusion we're trying to avoid? User asks us to install Ruby explicitly...and we say we assume they don't want Ruby installed.

cli/toolchain_cmd.go
```diff
@@ -538,7 +494,7 @@ func installJavaScriptToolchain() error {
 	srclibpathDir := filepath.Join(strings.Split(srclib.Path, ":")[0], toolchain) // toolchain dir under SRCLIBPATH
 
 	if _, err := exec.LookPath("node"); isExecErrNotFound(err) {
-		return skippedToolchain{toolchain, "no `node` in PATH (assuming you don't have Node.js installed and you don't want the JavaScript toolchain)"}
+		return errors.New("no `node` in PATH (assuming you don't have Node.js installed and you don't want the JavaScript toolchain)")
```
rothfels: Et al

Makefile
rothfels: @slimsag right, I meant "std" is kinda a deprecated term; this is more like `install-all-toolchains`

slimsag: Done.

cli/toolchain_cmd.go
slimsag: Done.

cli/toolchain_cmd.go
slimsag: Replaced with `(do you have LANGUAGE installed properly?)`

cli/toolchain_cmd.go
slimsag: done

renfredxh:
Remove (this line)[https://github.com/sourcegraph/srclib/blob/7e6ce0746ea6043833d3a229085dffc16c2c8c3c/docs/theme/home.html#L72] in the docs referencing skipping functionality

renfredxh:
Everything else LGTM

beyang:
SGTM

PR #199, opened 2015-09-10
store.defQueryTreeIndex: Prevent possible nil pointer dereference panic.

shurcooL:
There are situations where `x.mt.t` is nil, so doing `node, i := x.mt.t.IndexedTraverse([]rune(q))` results in a nil pointer dereference panic.

This is a quick fix to prevent a panic. A proper fix will be to avoid calling `getByQuery` when the mafsaTable is not built/read.

### DIFF
--- a/store/def_query_tree_index.go
+++ b/store/def_query_tree_index.go
@@ -43,6 +43,11 @@ func (x *defQueryTreeIndex) getByQuery(q string) (map[unit.ID2]byteOffsets, bool
 		panic("mafsaTable not built/read")
 	}
 
+	if x.mt.t == nil {
+		vlog.Println("getByQuery: x.mt.t == nil")
+		return nil, false
+	}
+
 	q = strings.ToLower(q)
 	node, i := x.mt.t.IndexedTraverse([]rune(q))
 	if node == nil {


store/def_query_tree_index.go
```diff
@@ -43,6 +43,11 @@ func (x *defQueryTreeIndex) getByQuery(q string) (map[unit.ID2]byteOffsets, bool
 		panic("mafsaTable not built/read")
 	}
```
shurcooL: I know there's a `if x.mt == nil {` guard just above, and it panics, so clearly this func is not meant to be called when `x.mt.t` is nil. That's why it's a short term fix.

store/def_query_tree_index.go
```diff
@@ -43,6 +43,11 @@ func (x *defQueryTreeIndex) getByQuery(q string) (map[unit.ID2]byteOffsets, bool
...
 
+	if x.mt.t == nil {
+		log.Println("getByQuery: x.mt.t == nil")
```
shurcooL: Hmm, this should probably be `vlog` to be consistent with the logging in this package?

neelance:
LGTM

store/def_query_tree_index.go
```diff
@@ -43,6 +43,11 @@ func (x *defQueryTreeIndex) getByQuery(q string) (map[unit.ID2]byteOffsets, bool
 		panic("mafsaTable not built/read")
 	}
 
+	if x.mt.t == nil {
+		vlog.Println("getByQuery: x.mt.t == nil")
```
shurcooL: This message will only show up if you use `V=t` env var.

PR #202, opened 2015-09-23
Allow `srclib toolchain install go` without GOPATH being set.

slimsag:
Also requires: https://github.com/sourcegraph/srclib-go/pull/49


beyang:
LGTM

PR #203, opened 2015-09-23
Make `srclib toolchain install go` actually install Go if it can.

slimsag:
This PR automatically installs Go itself if it is not already installed. Note that:

 - The user is prompted for "Do you want to install Go via <your_platform_method>? [y/n]:" otherwise installation aborts.
 - On Linux, it assumes that `apt-get` is available. If it is not, installation aborts.
 - On OS X, it assumes that `brew` is available. If it is not, installation aborts.

I've tested the command sets on both Ubuntu 14.04 and OS X.

### DIFF
--- a/cli/toolchain_cmd.go
+++ b/cli/toolchain_cmd.go
@@ -13,6 +13,7 @@ import (
 	"path"
 	"path/filepath"
 	"time"
+	"go/build"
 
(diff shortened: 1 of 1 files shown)


shurcooL:
Was this merged or abandoned? Looks like the latter.

PR #205, opened 2015-10-02
merge master into nodb

rothfels:
@beyang everything seems to be working and tested; once this is merged we should have an easy time merging nodb -> master.

### DIFF
--- a/toolchain/add.go
+++ b/toolchain/add.go
@@ -32,7 +32,8 @@ func Add(dir, toolchainPath string, opt *AddOpt) error {
 		return err
 	}
 
-	targetDir := filepath.Join(srclib.PathEntries()[0], toolchainPath)
+	srclibpathEntry := filepath.SplitList(srclib.Path)[0]
+	targetDir := filepath.Join(srclibpathEntry, toolchainPath)
 
 	if err := os.MkdirAll(filepath.Dir(targetDir), 0700); err != nil {
 		return err
--- a/toolchain/get.go
+++ b/toolchain/get.go
@@ -5,8 +5,6 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-
-	"sourcegraph.com/sourcegraph/srclib"
 )
 
 // CloneOrUpdate downloads the toolchain named by the toolchain path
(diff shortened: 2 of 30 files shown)


toolchain/get.go
```diff
@@ -15,21 +17,19 @@ import (
 func CloneOrUpdate(path string, update bool) (*Info, error) {
 	path = filepath.Clean(path)
 
-	dir, err := Dir(path)
-	if err != nil {
-		return nil, err
-	}
+	dir := filepath.SplitList(srclib.Path)[0]
```
beyang: This is incorrect. I think this should still use the `Dir` function in this package. Otherwise, instead of installing the toolchains in `~/.srclib/` (which is where they are supposed to be), this will install them into `~` (the parent of `$SRCLIBPATH`).

toolchain/add.go
```diff
@@ -32,7 +32,8 @@ func Add(dir, toolchainPath string, opt *AddOpt) error {
 		return err
 	}
 
-	targetDir := filepath.Join(srclib.PathEntries()[0], toolchainPath)
```
beyang: it looks like `PathEntries` is no longer used anywhere, so delete it.

beyang:
@rothfels the 2 comments above should be addressed, but otherwise LGTM!

toolchain/add.go
rothfels: Actually, it's being used by the `Dir` function:

```
func Dir(toolchainPath string) (string, error) {
	toolchainPath = filepath.Clean(toolchainPath)

	dir, err := lookupToolchain(toolchainPath)
	if os.IsNotExist(err) {
		return filepath.Join(srclib.PathEntries()[0], toolchainPath), nil
	}
	if err != nil {
		err = &os.PathError{Op: "toolchain.Dir", Path: toolchainPath, Err: err}
	}
	return dir, err
}
```

Should `srclib.PathEntries()[0]` in the definition above be replaced with `filepath.SplitList(srclib.Path)[0]` so we can remove `PathEntries`?

beyang: Yeah, that sounds good.

PR #212, opened 2015-10-19
cli: Fix double file close and remove unnecessary chmod.

jpoler:
The the file closing was deferred and then also closed
immediately. Changed to simply defer the close and log
it on failure.

Also removed an execution of chmod -R 777 on the repository
being configured. This isn't needed, when a volume is mounted,
the srclib user should be able to write to files without doing
this. Confirmed manually with the same src-preconfigcommands
Docker image. The user could nuke the entire filesystem if so
desired.


beyang:
Awesome, thanks for catching and fixing these!

PR #213, opened 2015-10-19
proto: re-generate protobufs with latest gogo/protobuf

pararthshah:
Updated github.com/gogo/protobuf to commit `200875` and did `go generate ./...`


shurcooL:
LGTM.

PR #215, opened 2015-10-27
toolchain: add support for Docker TLS (required for newer versions of Docker)

slimsag:
As of recent versions of Docker (e.g. boot2docker 1.3.0 and Docker Toolbox) only
TLS connections to the Docker daemon are allowed. This adds support for TLS so
that everything works out-of-the-box. Please see the code comments for additional
info.

### DIFF
--- a/toolchain/docker.go
+++ b/toolchain/docker.go
@@ -68,15 +68,14 @@ func newTLSDockerClient(certPath, host string) (*docker.Client, error) {
 // DOCKER_HOST env var, or the default /var/run/docker.sock socket if unset.
 func newDockerClient() (*docker.Client, error) {
 	dockerEndpoint := os.Getenv("DOCKER_HOST")
-	if dockerEndpoint == "" {
-		dockerEndpoint = "unix:///var/run/docker.sock"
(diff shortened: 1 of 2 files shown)


toolchain/docker.go
```diff
@@ -15,6 +72,11 @@ func newDockerClient() (*docker.Client, error) {
 		dockerEndpoint = "unix:///var/run/docker.sock"
 	} else if !strings.HasPrefix(dockerEndpoint, "http") && !strings.HasPrefix(dockerEndpoint, "tcp") {
 		dockerEndpoint = "http://" + dockerEndpoint
+	} else if strings.HasPrefix(dockerEndpoint, "tcp") {
+		certPath := os.Getenv("DOCKER_CERT_PATH")
+		if certPath != "" {
```
beyang: `if certPath := os.Getenv("DOCKER_CERT_PATH"); certPath != "" {` is more concise.

beyang: If this is the default in the newer Docker versions, perhaps we should move this `if` statement up in precedence so that it is preferred above other connection methods, and the others are viewed as fallbacks?

slimsag: Agreed on both accounts.

slimsag:
I've addressed both of your comments + updated the godep to fix the Travis CI build, PTAL.

PR #217, opened 2015-10-27
cli: revert a part of PR #212

slimsag:
This change reverts a part of PR #212 / 3dcee737b542f81af862a93449eb9deb855cce3f which
removed a hack on the pretense that it worked fine locally without it. Although it is
true that it works fine without it locally (e.g. running `srclib config -m docker` works),
we are seeing failures to build the github.com/golang/go repository due to file permission
errors:

```
##### Building Go bootstrap tool. cmd/dist
# _/src/src/cmd/dist cannot create cmd/dist/dist: Permission denied
PreConfigCommands: command "\nif [ -d /home/srclib ]; then\ncurl -L 'https://storage.googleapis.com/golang/go1.4.1.linux-amd64.tar.gz' > /tmp/go1.4.tgz &&\nmkdir /home/srclib/go1.4 &&\ntar -xf /tmp/go1.4.tgz -C /home/srclib/go1.4 --strip-components=1\nfi &&\ncd src\n./make.bash\n": exit status 2
```

Suspected is that older Docker versions handle the file permissions differently, thus it works
locally but not in prod.

### DIFF
--- a/cli/config_cmd.go
+++ b/cli/config_cmd.go
@@ -240,6 +240,15 @@ WORKDIR /src
 			return err
 		}
 
+		// HACK: make the files writable by the docker user
+		//
+		// It is suspected that this is needed due to some older docker versions. We
+		// should investigate why it is needed for building e.g. the github.com/golang/go
+		// repository and remove this hack when possible.
+		if err := exec.Command("chmod", "-R", "777", dir).Run(); err != nil {
+			return fmt.Errorf("chmod -R 777 dir failed: %s", err)
+		}
+
 		for _, cmdStr := range cmds {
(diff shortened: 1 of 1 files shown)


slimsag:
It would be nice to be able to verify that this is indeed the issue before merging it back in, since it's clearly marked as a HACK. Unfortunately I cannot reproduce the issue locally.

beyang:
It's a harmless hack, so let's just try it.

slimsag:
Good news: I've managed to confirm that this does indeed fix the issue on our workers (e.g. `Docker version 1.6.2, build 7c8fca2`).

PR #219, opened 2015-11-12
Fix counter data race

slimsag:
This change fixes many data races present in the tests, where the actual method being tested runs several other goroutines, with each trying to touch a global counter variable. After this change, I am able to successfully run all tests with the race detector enabled.

Also enable the Go race detector in Travis, to help catch these in the future.

### DIFF
--- a/store/counter.go
+++ b/store/counter.go
@@ -1,31 +1,23 @@
 package store
 
-import "sync"
+import "sync/atomic"
 
 // counter is a simple thread-safe integer.
 type counter struct {
-	sync.RWMutex
-	count int
+	count int64
 }
 
 // increment increments the counter by one.
-func (c *counter) increment() {
-	c.Lock()
-	c.count++
-	c.Unlock()
+func (c counter) increment() {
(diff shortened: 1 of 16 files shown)


store/counter.go
```diff
@@ -0,0 +1,31 @@
...
+// counter is a simple thread-safe integer.
+type counter struct {
+	sync.RWMutex
+	count int
+}
+
+// increment increments the counter by one.
+func (c *counter) increment() {
+	c.Lock()
+	c.count++
```
keegancsmith: you can skip out on the locking, and use the atomic package to do the increments.

keegancsmith:
Looks good to me! Also adding race detection CI ftw.

store/counter.go
shurcooL: Agreed, this looks like a good situation to consider using atomic increments instead of locking.

shurcooL:
This was merged as https://github.com/sourcegraph/srclib/commit/ff4f690b157ce154af2f44018141c49b2d5b17df.

PR #222, opened 2015-11-17
Improvements for docker mode

alexsaveliev:
- removed hack for srclib-java in Docker mode
- returning error instead of panicking when we were unable to normalize repo clone URI

### DIFF
--- a/cli/test_cmd.go
+++ b/cli/test_cmd.go
@@ -115,8 +115,7 @@ func (c *TestCmd) Execute(args []string) error {
 			expectedDir := filepath.Join(tree, "../../expected", exeMethod, filepath.Base(tree))
 			actualDir := filepath.Join(tree, "../../actual", exeMethod, filepath.Base(tree))
 
-			_, err := os.Stat(expectedDir)
-			if err != nil && os.IsNotExist(err) {
+			if _, err := os.Stat(expectedDir); err != nil && os.IsNotExist(err) {
 				// (alexsaveliev) we may have some tests available only for docker or program mode
 				if GlobalOpt.Verbose {
 					log.Printf("Skipping tree %v...", tree)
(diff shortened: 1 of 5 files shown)


cli/test_cmd.go
```diff
@@ -114,6 +114,16 @@ func (c *TestCmd) Execute(args []string) error {
 			}
 			expectedDir := filepath.Join(tree, "../../expected", exeMethod, filepath.Base(tree))
 			actualDir := filepath.Join(tree, "../../actual", exeMethod, filepath.Base(tree))
+
+			_, err := os.Stat(expectedDir)
+			if err != nil && os.IsNotExist(err) {
```
beyang: `if _, err := os.Stat(expectedDir); err != nil && os.IsNotExist(err) {`

cli/test_cmd.go
```diff
@@ -114,6 +114,16 @@ func (c *TestCmd) Execute(args []string) error {
...
+				// (alexsaveliev) we may have some tests available only for docker or program mode
```
beyang: Why would a test only exist for docker or program mode and not both? The two should have equivalent behavior (as much as practically possible).

cli/test_cmd.go
alexsaveliev: fixed

cli/test_cmd.go
alexsaveliev: Right, they should behave identical as much as possible, but what Docker image has some pre-requisites missing (e.g. Android SDK)?

cli/test_cmd.go
```diff
@@ -114,6 +114,15 @@ func (c *TestCmd) Execute(args []string) error {
 			}
 			expectedDir := filepath.Join(tree, "../../expected", exeMethod, filepath.Base(tree))
 			actualDir := filepath.Join(tree, "../../actual", exeMethod, filepath.Base(tree))
+
+			if _, err := os.Stat(expectedDir); err != nil && os.IsNotExist(err) {
```
beyang: `if _, err := os.Stat(expectedDir); os.IsNotExist(err) {`

beyang: Actually, can we just remove this clause, and generate expected output for both the program and Docker cases in srclib-java?

beyang:
after comment is addressed, LGTM.

cli/test_cmd.go
alexsaveliev: Reverted changes related to abliity to exclude missing tests

PR #223, opened 2015-11-26
Tolerating missing files

alexsaveliev:
- When file is present in unit's Files section but it's absent on disk, we should exclude it from Makefile prerequisites instead of terminating program execution. This situation may occur for example when file is being generated during compile phase if toolchain runs in the Docker mode: generated source file may be detected and included into Files, but after Docker container is stopped, file is no longer accessible.

### DIFF
--- a/grapher/rule.go
+++ b/grapher/rule.go
@@ -3,6 +3,7 @@ package grapher
 import (
 	"fmt"
 	"log"
+	"os"
 	"path/filepath"
 
 	"sourcegraph.com/sourcegraph/makex"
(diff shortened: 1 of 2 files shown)


grapher/rule.go
```diff
@@ -58,7 +59,13 @@ func (r *GraphUnitRule) Target() string {
 
 func (r *GraphUnitRule) Prereqs() []string {
 	ps := []string{filepath.ToSlash(filepath.Join(r.dataDir, plan.SourceUnitDataFilename(unit.SourceUnit{}, r.Unit)))}
-	ps = append(ps, r.Unit.Files...)
+	for _, file := range r.Unit.Files {
+		if _, err := os.Stat(file); err != nil && os.IsNotExist(err) {
```
shurcooL: @beyang made a suggestion to use just `os.IsNotExist(err)` (i.e., drop the `err != nil`) in another PR with similar code, but not here.

Is there a reason for that? Which style was deemed to be preferable?

beyang: I think I just overlooked this. Just using `os.IsNotExist` is more concise.

PR #224, opened 2015-12-09
docker: Use --rm for run

keegancsmith:
What we run is ephemeral, but docker by default leaves the container around even
once it has exited. This is causing issues in production since we have 10s of
thousands instances in `docker ps -a`, which seems to be causing perf issues in
docker.


pararthshah:
LGTM

PR #225, opened 2015-12-10
doc: Return template.HTML in toHTML

MartinAmps:


beyang:
Will this change break any of the code in the Sourcegraph main repository?

MartinAmps:
Yes, it needs to be coordinated with this CS (almost finished): https://src.sourcegraph.com/sourcegraph/.changes/420

slimsag:
I haven't had a chance to look into this further, but actually @MartinAmps I'm a bit confused about the relation of this `doc/doc.go` and the one in the main repo. They appear to be pretty similar (identical at one point?) but are now not?

It appears that this ToHTML does not sanitize the output, so returning `template.HTML` here is a bit falsey. We need to look into where this doc package is actually used / if this change is needed, and I also wonder if they can be deduplicated (maybe out of scope of this change)?

shurcooL:
> returning `template.HTML` here is a bit falsey

Agreed. `template.HTML` should indeed be reserved for sanitized HTML that is safe to forward to end client (typically, user's browser) without further processing.

> HTML encapsulates a known safe HTML document fragment. It should not be used for HTML from a third-party, or HTML with unclosed tags or comments. The outputs of a sound HTML sanitizer and a template escaped by this package are fine for use with HTML.

https://godoc.org/html/template#HTML

It's not the case here (namely, output of `blackfriday.MarkdownCommon` is not sanitized and not safe to forward to the user's browser; it may contain scripts and other html tags).

MartinAmps:
We can replace the use of srclib/doc with sourcegraph/doc which is sanitized appropriately to still return a byte array here. Sorry for the confusion!

PR #226, opened 2015-12-13
update import path for rog-go

sqs:
`code.google.com/p/rog-go -> github.com/rogpeppe/rog-go`

This is to eliminate the need for `hg` when building srclib. This is the last `hg` dep.


shurcooL:
:confetti_ball:

PR #227, opened 2015-12-14
store: remove def files index

sqs:
The def files index was hacky: it indexed the first 7 defs underneath
any path. It was tied to the presentation needs of Sourcegraph. This
was confusing. Now that Sourcegraph no longer even needs that
presentation mode, we can remove it.


shurcooL:
Great! It's always nice when code can be removed (while preserving functionality).

keegancsmith:
This is great \o/
There are some unrelated failures on CI. They look pretty easy to fix, do you mind fixing before pushing?

PR #228, opened 2015-12-14
Adding srclib-basic toolchain

alexsaveliev:
This pull request adds srclib-basic toolchain and provides temporary workaround for tools/godep#359


beyang:
Here are some PHP issues we should address before merging and deploying this.

On github.com/PHPMailer/PHPMailer:
image

- definition highlighting and reference highlighting misaligned
- page load takes 8.6 seconds

image

- reference highlighting is incorrect

Is this because the language-independent syntax highlighting lumps the reference tokens together when they should be separate? Would adding PHP support to syntaxhighlight fix this?

alexsaveliev:
> definition highlighting and reference highlighting misaligned

Claimed fixed (https://src.sourcegraph.com/syntaxhighlight/.changes/1)

> page load takes 8.6 seconds

Performance improved. PHP lexer have no more RE-based rules, all of them were replaced with functions

> reference highlighting is incorrect

Claimed fixed (https://src.sourcegraph.com/syntaxhighlight/.changes/1). The problem was that PHP lexer expected full PHP source and was unable to highlight properly PHP fragments

beyang:
Thanks for fixing these. Additional bugs I discovered using it some more:

- Go to http://localhost:3080/github.com/PHPMailer/PHPMailer@master/.tree/extras/EasyPeasyICS.php?startline=133&endline=141&seldef=/github.com/PHPMailer/PHPMailer@master/.basic-php/php/.def/:EasyPeasyICS/$calendarName. The last example has incorrect syntax highlighting:
image

beyang:
What happens in the case where a reference's definition is ambiguous/uncertain? Do we just pick and choose one definition in that case?

Sourcegraph supports having multiple overlapping tokens (useful for some language features, when a token can reference simultaneously a field and a type like in Go). We could use that here to let the user choose between possible choices when the definition choice is uncertain.

PR #229, opened 2015-12-21
Remove first-class support for execution inside Docker

sqs:
Remove the first-class support for the Docker execution method. You can still run srclib inside Docker, but it is not natively Docker-aware. It's still a good idea for individual toolchains to provide Dockerfiles for ease of development.

Outstanding TODOs
* Remove `toolchain.Add` func


cli/toolchain_cmd.go
```diff
@@ -86,15 +67,6 @@ func init() {
 		log.Fatal(err)
 	}
 
-	_, err = c.AddCommand("add",
```
alexsaveliev: Sorry, but what is the reason of sacrificing `toolchain add`? When working on new (or forked) toolchain I'm using `add` a lot..

sqs: It is confusing to the 97% of users that aren't srclib toolchain developers. Several people have mentioned it. I think all it does is create a symlink. Are you OK with just manually creating the symlink as needed?

Other option is to rename it something like `link-dev-toolchain`, but IMO I prefer when simple operations like symlinks are exposed to the user instead of wrapped in an alias.

alexsaveliev: Got it, thanks. Will use symlinks

shurcooL:
What's the status of this PR being merged to `master`? It's already being used in Sourcegraph via vendoring.

PR #235, opened 2015-12-31
cli: expose AddCommands to allow mounting srclib as a subcommand of another CLI

sqs:
This lets you create another program "myprogram" that mounts srclib's CLI as "myprogram srclib ...". For example, "myprogram srclib toolchain list" would run that subcommand of srclib.

Mostly an internal change only likely to affect (and benefit) Sourcegraph.


sqs:
Please also review https://src.sourcegraph.com/sourcegraph/.changes/510

PR #236, opened 2016-01-05
{store,cli}: add specific option (no prefixes) to ByFiles

slimsag:
This change adds a `exact` flag to the `ByFiles` filter such that it will
not be inclusive of files matching the prefix (e.g. `foo/` does not match
`foo/bar.go`).

### DIFF
--- a/store/filter.go
+++ b/store/filter.go
@@ -702,10 +702,9 @@ type ByFilesFilter interface {
 // empty, or if the file path has not been cleaned (i.e., if file !=
 // path.Clean(file)).
 //
-// If specific == true, then only the specific files are accepted (i.e. a
-// directory path will not include all files in that directory as it would
-// normally).
-func ByFiles(specific bool, files ...string) interface {
+// If exact == true, then only the exact files are accepted (i.e. a directory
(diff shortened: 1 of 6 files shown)


store/filter.go
```diff
@@ -701,7 +701,11 @@ type ByFilesFilter interface {
 // or contain any of the listed files. It panics if any file path is
 // empty, or if the file path has not been cleaned (i.e., if file !=
 // path.Clean(file)).
-func ByFiles(files ...string) interface {
+//
+// If specific == true, then only the specific files are accepted (i.e. a
+// directory path will not include all files in that directory as it would
+// normally).
+func ByFiles(specific bool, files ...string) interface {
```
beyang: `specific` doesn't immediately convey to me the meaning of the parameter. What about `exact`, which I've seen more commonly in our codebase and others, or `prefix` (if you want to flip the meaning of the parameter)?

slimsag: Good idea -- I had trouble coming up with the correct name for this parameter. I do like `exact` much more.

beyang: LGTM.

PR #237, opened 2016-01-06
2.0 schema

beyang:
This changeset defines the data structures in the srclib v2 schema and the `config2` and `make2` commands.

Because the structure of execution did not change that much, but the data structures being passed around did, it was easiest to copy-and-modify the existing commands to use the new data structures instead of the v1 structures.

A good place to start the review is in the `graph2/graph.proto` file, which contains most of the new core structures.

This preserves the old schema for now, so should not break any existing srclib functionality.


graph2/graph.proto
```diff
@@ -0,0 +1,307 @@
...
+message RepoPermissions {
+	bool read = 1;
+	bool write = 2;
+	bool admin = 3;
+}
+
+message TreeKey {
+  // TreeType is the type of source tree.
+  // (git, hg, build-system specific like pip or npm, ftp)
+  string tree_type = 1;
```
shurcooL: Minor. Leading whitespace is inconsistent. It's a tab above in `GitHubRepo` and `RepoPermissions`, but 2 spaces here and below. It would be better to keep a consistent style. I'd be in favor of staying close to gofmt and using a single tab for indentation (but spaces for alignment).

beyang: Thanks for the comment. I think my editor configuration was different on my mac. Will fix before merging.

sqs:
Can you merge no-docker first to make this diff only contain relevant changes?

graph2/graph.proto
```diff
@@ -0,0 +1,307 @@
...
+	// source trees of the same type. It also typically encodes how the
+	// source tree should be fetched.
+	string uri = 2 [(gogoproto.customname) = "URI"];
+}
+
+// Tree is a source tree, the outermost container of code in the
+// srclib schema. Examples include VCS repositories and package
+// manager packages/modules. A source tree contains zero or more build
+// units.
+message Tree {
```
sqs: Why are repo metadata items in here? Description, GitHub star count, language, private, etc.

graph2/graph.proto
```diff
@@ -0,0 +1,307 @@
...
+
+message TreeKey {
+	// TreeType is the type of source tree.
+	// (git, hg, build-system specific like pip or npm, ftp)
+	string tree_type = 1;
+
+	// URI uniquely identifies a source tree among the universe of
+	// source trees of the same type. It also typically encodes how the
+	// source tree should be fetched.
+	string uri = 2 [(gogoproto.customname) = "URI"];
```
sqs: For consistency with sourcegraph now, make these upper-camel-cased (URI not `uri`, TreeType not `tree_type`).

graph2/graph.proto
```diff
@@ -0,0 +1,307 @@
...
+option (gogoproto.goproto_getters_all) = false;
+
+// GitHubRepo holds additional metadata about GitHub repos.
+message GitHubRepo {
+	int32 stars = 1;
+}
+
+// RepoPermissions describes the possible permissions that a user (or an anonymous
+// user) can be granted to a repository.
+message RepoPermissions {
```
sqs: Why is RepoPermissions a srclib concern?

graph2/graph.proto
```diff
@@ -0,0 +1,307 @@
...
+
+message TreeRev {
+	Tree tree = 1;
+
+	// Rev is the version of the source tree
+	string rev = 2;
+
+	// Type of revision (e.g., commit ID, branch name, or tag for VCS
+	// repositories; version string for build system packages)
+	string rev_type = 3;
```
sqs: What are the possible values here? The strings "commit", "branch", "tag", and "version string"? Should this be an enum or just fully documented?

graph2/graph.proto
```diff
@@ -0,0 +1,307 @@
...
+	// if there is no current user) is granted to this repository.
+	RepoPermissions permissions = 18;
+
+	GitHubRepo github = 19 [(gogoproto.customname) = "GitHub"];
+}
+
+message TreeRev {
+	Tree tree = 1;
+
+	// Rev is the version of the source tree
```
sqs: Use a period at the end of the sentence for consistency.

graph2/graph.proto
beyang: Mainly for backcompat. The information is currently being used, so it needs to be stored somewhere.

graph2/graph.proto
```diff
@@ -0,0 +1,307 @@
...
+	string version = 2;
+}
+
+// UnitKey universally uniquely identifies a build unit.
+message UnitKey {
+	TreeKey key = 1 [(gogoproto.nullable) = false, (gogoproto.embed) = true, (gogoproto.jsontag) = ""];
+
+	// Version is the build-system-level version string. It can also be a commit
+	// ID, but no client code should depend on implied VCS semantics (e.g., never
+	// run `git clone buildUnit.Version`).
```
sqs: The `git clone` command example doesn't make sense to me - why would someone try to run `git clone` on something that isn't a clone URL anyway?

graph2/graph.proto
```diff
@@ -0,0 +1,307 @@
...
+	// unique identifier within a given source unit. In many languages, it is
+	// convenient to use the namespace hierarchy (with some modifications) as
+	// the Path, but this may not always be the case. I.e., don't rely on Path
+	// to find parents or children or any other structural propreties of the
+	// def hierarchy). See Def.TreePath instead.
+	string path = 2 [(gogoproto.jsontag) = "Path"];
+}
+
+// Node is a node in the code graph extracted from the source code.
+message Node {
```
sqs: Give examples.

graph2/graph.proto
```diff
@@ -0,0 +1,307 @@
...
+	// formats include 'text/html', 'text/plain', 'text/x-markdown', text/x-rst'.
+	string format = 1;
+
+	// Data is the actual documentation text.
+	string data = 2;
+}
+
+// Edge describes the relationship between two nodes in the source graph.
+// These can extend across build units.
+message Edge {
```
sqs: Give examples of what an edge could be.

graph2/graph.proto
```diff
@@ -0,0 +1,307 @@
...
+	// the Path, but this may not always be the case. I.e., don't rely on Path
+	// to find parents or children or any other structural propreties of the
+	// def hierarchy). See Def.TreePath instead.
+	string path = 2 [(gogoproto.jsontag) = "Path"];
+}
+
+// Node is a node in the code graph extracted from the source code.
+message Node {
+	NodeKey key = 1 [(gogoproto.nullable) = false, (gogoproto.embed) = true, (gogoproto.jsontag) = ""];
+	string kind = 2;
```
sqs: What the the possible values for `kind`?

graph2/graph.proto
sqs: URI means something different IMO. I think it's misused here. Why not just call it an ID? How is it different from something you'd call an ID?

graph2/graph.proto
```diff
@@ -0,0 +1,307 @@
...
+message RepoPermissions {
+	bool read = 1;
+	bool write = 2;
+	bool admin = 3;
+}
+
+message TreeKey {
+	// TreeType is the type of source tree.
+	// (git, hg, build-system specific like pip or npm, ftp)
+	string tree_type = 1;
```
sqs: Why `TreeKey.TreeType` and not just `TreeKey.Type`?

graph2/graph.proto
```diff
@@ -0,0 +1,307 @@
...
+package graph2;
+
+import "github.com/gogo/protobuf/gogoproto/gogo.proto";
+import "sourcegraph.com/sqs/pbtypes/timestamp.proto";
+import "sourcegraph.com/sourcegraph/srclib/toolref.proto";
+
+option (gogoproto.goproto_getters_all) = false;
+
+// GitHubRepo holds additional metadata about GitHub repos.
+message GitHubRepo {
```
sqs: As noted below, why does srclib care about GitHub repo star counts?

graph2/graph.proto
beyang: It will typically be a URI/URL-like thing, like "github.com/gorilla/mux" or "hg.python.org/cpython" or "pypi.python.org/urllib3". Are there non-URI IDs you're thinking of?

graph2/graph.proto
beyang: good point, copy-pasta, will remove

PR #239, opened 2016-01-15
changed protobuf field names to camel case

neelance:


neelance:
These changes were performed by the https://src.sourcegraph.com/sourcegraph@master/.tree/dev/fix-proto.go tool.

PR #241, opened 2016-01-28
Handling UTF-8 BOM and adjusting srclib-csharp offsets

alexsaveliev:
- Bumping version of `github.com/sqs/fileset` to include recent changes related to UTF-8 BOM support
- `srclib-csharp` emits defs and refs character-based offsets so they need to be converted to byte ones.

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -70,7 +70,7 @@
 		},
 		{
 			"ImportPath": "github.com/sqs/fileset",
-			"Rev": "4317e899aa9438ba7603a6e322389571cb3ffdff"
+			"Rev": "af79d58d268e6fb30711babccd37156c95fd5aaf"
 		},
 		{
 			"ImportPath": "github.com/sqs/go-selfupdate/selfupdate",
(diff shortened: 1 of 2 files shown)


shurcooL:
> - Bumping version of `github.com/sqs/fileset` to include recent changes related to UTF-8 BOM support

That part doesn't seem right. Are you sure it's done correctly?

I'm seeing that only the `Godeps/Godeps.json` file has changed, but none of the actual vendored dependency code has changed.

Given the diff in commits, I'd expect to see this change:

https://github.com/sqs/fileset/compare/4317e899aa9438ba7603a6e322389571cb3ffdff...af79d58d268e6fb30711babccd37156c95fd5aaf

(Maybe not include the _test.go file.)

shurcooL:
Ah, I see that this repository uses a non-copy godep mode. It only has the Godeps.json file, but doesn't copy/vendor the actual dependencies inside the repo. So the change above is valid.

Sorry for the noise, and thanks for pointing it out @alexsaveliev!

Issue #242, opened 2016-01-29
SCM URL support


alexsaveliev:
I think that srclib should be able to parse the following URL:

- scm:VCS:PROTOCOL.... (example: scm:git:https://server_name[:port]/path_to_repository)
- VCS:PROTOCOL (example: git:https://server_name[:port]/path_to_repository)

Please see more samples here: https://confluence.jetbrains.com/display/TCD9/Guess+Settings+from+Repository+URL

Current implementation 
```go
func TryMakeURI(cloneURL string) (string, error)
``` 

in `repo.go` doesn't

PR #244, opened 2016-01-30
Fixes sourcegraph/srclib#242

alexsaveliev:
- adding SCM URL handling, now we should support the following URLs:
  - protocol:... (https://foo.bar)
  - vcs:... (git://foo.bar)
  - vcs:protocol:.... (git:https://foo.bar)
  - scm:vcs:protocol:... (scm:hg:https://foo.bar)
  - host:path (assumed SSH)

### DIFF
--- a/graph/repo.go
+++ b/graph/repo.go
@@ -37,9 +37,7 @@ func TryMakeURI(cloneURL string) (string, error) {
 	}
 
 	// Removing leading "scm:" if any
-	if strings.HasPrefix(cloneURL, "scm:") {
-		cloneURL = cloneURL[4:]
-	}
+	cloneURL = strings.TrimPrefix(cloneURL, "scm:")
 
 	// Removing VCS part if any, git:http://.. => http://..
 	cloneURL = removeVCSPart(cloneURL)
@@ -74,7 +72,7 @@ func URIEqual(a, b string) bool {
 	return strings.EqualFold(a, b)
 }
 
-// Removes VCS part if any, git:http://.. => http://..
+// removeVCSPart removes VCS part from URL if any, git:http://.. => http://..
 func removeVCSPart(url string) string {
 	parts := strings.SplitN(url, ":", 3)
 	if len(parts) < 3 {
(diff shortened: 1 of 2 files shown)


graph/repo.go
```diff
@@ -25,11 +25,25 @@ func MakeURI(cloneURL string) string {
...
+// - host:path (assumed SSH)
 func TryMakeURI(cloneURL string) (string, error) {
 	if cloneURL == "" {
 		return "", errors.New("MakeURI: empty clone URL")
 	}
 
+	// Removing leading "scm:" if any
+	if strings.HasPrefix(cloneURL, "scm:") {
+		cloneURL = cloneURL[4:]
+	}
```
shurcooL: Suggestion:

```Go
cloneURL = strings.TrimPrefix(cloneURL, "scm:")
```

graph/repo.go
```diff
@@ -59,3 +73,28 @@ func TryMakeURI(cloneURL string) (string, error) {
...
+		if !isASCIILetter(r) {
+			return url
+		}
+	}
+	return parts[1] + ":" + parts[2]
+}
+
+// Returns true if given rune is ASCII letter
+func isASCIILetter(r rune) bool {
+	return r >= 'a' && r <= 'z' || r >= 'A' && r <= 'Z'
```
shurcooL: Really minor style thought, but I [would write](https://github.com/gopherjs/snippet-store/blob/1f77889ea9974c582661da570ae5c35686763b10/id.go#L31) this as:

```Go
return 'a' <= r && r <= 'z' || 'A' <= r && r <= 'Z'
```

It's up to you which you think is better.

graph/repo.go
```diff
@@ -59,3 +73,28 @@ func TryMakeURI(cloneURL string) (string, error) {
...
+	for _, r := range parts[1] {
+		if !isASCIILetter(r) {
+			return url
+		}
+	}
+	return parts[1] + ":" + parts[2]
+}
+
+// Returns true if given rune is ASCII letter
+func isASCIILetter(r rune) bool {
```
shurcooL: Go style suggests the comment should begin with "isASCIILetter ...".

It's also a good practice (set by Rob Pike) to use the verb "reports" rather than the phrase "returns true" for funcs that return a single `bool` type.

So, this should be:

```Go
// isASCIILetter reports if given rune is an ASCII letter.
```

alexsaveliev:
@shurcooL, thanks a lot for your feedback. Your comments were taken into account

PR #246, opened 2016-02-03
Coverage command

jpoler:
@keegancsmith do you mind giving this a sanity check?

### DIFF
--- a/cli/coverage_cmd.go
+++ b/cli/coverage_cmd.go
@@ -11,6 +11,8 @@ import (
 	"path/filepath"
 	"strings"
 
+	"golang.org/x/tools/go/loader"
+
 	"github.com/kr/fs"
 
 	"sourcegraph.com/sourcegraph/go-flags"
@@ -32,6 +34,8 @@ func init() {
 var cacheDir = ".srclib-cache"
 var gitDir = ".git"
 var unitFile = "GoPackage.unit.json"
+var depFile = "GoPackage.depresolve.json"
+var graphFile = "GoPackage.graph.json"
 
 type CoverageCmd struct {
 	w io.Writer
@@ -39,11 +43,31 @@ type CoverageCmd struct {
 
 type Coverage struct {
 	Repo     *Repo
-	Warnings []string
+	Warnings []BuildWarning
 }
+
+type BuildWarning struct {
+	Directory string
+	Warning   WarningType
+}
+
+type WarningType string
+
+const (
+	BuildSucceededSrclibFailed = "Build succeeded but Srclib outputs failed"
+	BuildFailedSrclibSucceeded = "Build failed but Srclib succeeded"
+)
 
 var coverageCmd CoverageCmd
 
+// Execute performs a sanity check on srclib builds for go repositories.
+// We iterate every directory and do a coverage check on every package.
+// The coverage heuristic is very rough currently, but makes the following assumptions:
+// - only golang coverage is checked
+// - if a directory can be imported (see build standard library package) and built
+//   (see loader standard library package), then there should be three files present
+//   in the corresponding directory under .srclib-cache: a unit file, a depresolve file,
+//   and a graph file.
 func (c *CoverageCmd) Execute(args []string) error {
 	if c.w == nil {
 		c.w = os.Stdout
@@ -62,7 +86,7 @@ func (c *CoverageCmd) Execute(args []string) error {
 	var importPath string
 	for _, p := range strings.Split(goPath, ":") {
 		if strings.Contains(lRepo.RootDir, p) {
-			importPath = strings.TrimPrefix(lRepo.RootDir, filepath.Join(p, "src"))
+			importPath = strings.TrimPrefix(lRepo.RootDir, filepath.Join(p, "src")+"/")
 		}
 	}
 
@@ -83,8 +107,8 @@ func (c *CoverageCmd) Execute(args []string) error {
 	walker := fs.Walk(lRepo.RootDir)
 
 	for walker.Step() {
+
 		if err := walker.Err(); err != nil {
-			// should we output the current coverage anyway? (aka just break)
 			return err
 		}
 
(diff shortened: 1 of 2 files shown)


cli/coverage_cmd.go
```diff
@@ -0,0 +1,174 @@
...
+		return lRepoErr
+	}
+
+	goPath := os.Getenv("GOPATH")
+	if goPath == "" {
+		return errors.New("GOPATH not set")
+	}
+
+	var importPath string
+	for _, p := range strings.Split(goPath, ":") {
```
alexsaveliev: On Windows GOPATH will be separated by ";"

cli/coverage_cmd.go
```diff
@@ -0,0 +1,174 @@
...
+		if strings.Contains(lRepo.RootDir, p) {
+			importPath = strings.TrimPrefix(lRepo.RootDir, filepath.Join(p, "src")+"/")
```
alexsaveliev: There might be problem on Windows if RootDir looks Windows-style: `C:\foo\bar`

jpoler: @alexsaveliev thanks for taking a look. I don't have a windows machine, would you be able to confirm this?

cli/coverage_cmd.go
jpoler: Good point, will add.

cli/coverage_cmd.go
```diff
@@ -0,0 +1,179 @@
...
+	"github.com/kr/fs"
+
+	"sourcegraph.com/sourcegraph/go-flags"
+)
+
+func init() {
+	cliInit = append(cliInit, func(cli *flags.Command) {
+		_, err := cli.AddCommand("coverage",
+			"use a simple heuristic to check that srclib is outputting expected graph data",
+			`The coverage command acts as a sanity check to ensure that the srclib-go toolchain succeeds when it should and fails when it should not.`,
```
keegancsmith: Can we not explicitly mention `srclib-go`, since this is the more generic srclib project here.

cli/coverage_cmd.go
```diff
@@ -0,0 +1,179 @@
...
+
+	var importPath string
+	splitGoPath := filepath.SplitList(goPath)
+
+	for _, p := range splitGoPath {
+		p = filepath.ToSlash(p)
+		if strings.Contains(lRepo.RootDir, p) {
+			importPath = strings.TrimPrefix(lRepo.RootDir, filepath.Join(p, "src")+"/")
+		}
+	}
```
keegancsmith: I don't know srclib well enough to tell if this is a fragile way to find the package name. But given that this command has a very specific use case, lets ignore that for now and see if it works on our CI system.

cli/coverage_cmd.go
keegancsmith: Maybe mention that this isn't meant to usually be used by end users.

cli/coverage_cmd.go
```diff
@@ -0,0 +1,179 @@
...
+	"golang.org/x/tools/go/loader"
+
+	"github.com/kr/fs"
+
+	"sourcegraph.com/sourcegraph/go-flags"
+)
+
+func init() {
+	cliInit = append(cliInit, func(cli *flags.Command) {
+		_, err := cli.AddCommand("coverage",
```
keegancsmith: A better name might be sanity or validate? Coverage is a little overloaded due to the meaning it has from test coverage.

cli/coverage_cmd.go
```diff
@@ -0,0 +1,179 @@
...
+		}
+
+		pth := walker.Path()
+
+		fi, err := os.Stat(pth)
+		if !fi.IsDir() {
+			continue
+		}
+
+		if strings.Contains(pth, cacheDir) || strings.Contains(pth, gitDir) {
```
keegancsmith: we could just ignore anything that starts with a `.`

cli/coverage_cmd.go
```diff
@@ -0,0 +1,179 @@
...
+			})
+		} else if !importAndBuildSucceeded && srclibOutputsExist {
+			cov.Warnings = append(cov.Warnings, BuildWarning{
+				Directory: pth,
+				Warning:   BuildFailedSrclibSucceeded,
+			})
+		}
+	}
+
+	enc := json.NewEncoder(c.w)
```
keegancsmith: something else that would be useful in the json output would be some aggregate statistics on the warnings. Like number of Warnings, and a total for each warning type?

keegancsmith:
A few suggestions. Once addressed, lets maybe experiment with this on Sourcegraph before merging it into master.

FYI I think it is fine that this is super go specific right now and lives inside of srclib (rather than `srclib-go`). As we expand this to other langs, we can move code round into a slightly more complicated architecture.

cli/coverage_cmd.go
jpoler: sure, I'll go with `srclib validate`

cli/coverage_cmd.go
jpoler: makes sense

cli/coverage_cmd.go
jpoler: Good idea.

cli/coverage_cmd.go
jpoler: Also a good idea.

jpoler:
@keegancsmith Do you mind looking over this one more time? I've addressed everything but the aggregate stats. We can come back to that.

cli/validate_cmd.go
```diff
@@ -0,0 +1,178 @@
...
+		pth := walker.Path()
+
+		fi, err := os.Stat(pth)
+		if !fi.IsDir() {
+			continue
+		}
+
+		for _, part := range strings.Split(pth, string(os.PathSeparator)) {
+			if strings.HasPrefix(part, ".") {
+				fmt.Println("Skipping", pth)
```
keegancsmith: maybe use log here? I worry about outputting info like this to stdout + the json encoded line we actually care about.

keegancsmith:
@jpoler one inline comment. Otherwise LGTM

cli/validate_cmd.go
jpoler: good catch, I was only using that for debugging.

PR #248, opened 2016-02-15
robust file-not-found check

beyang:
Sometimes, `buildDataFS` is a virtual FS (e.g., zipfs), in which case `os.IsNotExist(err) == false`.

This fixes an issue that causes the import step to fail completely in Sourcegraph if any file is not found (which happens frequently on larger repositories, where srclib might fail on some source units).


sqs:
Indeed: https://github.com/golang/tools/blob/608d57b3ae930138a65e85b64edf2ba1b3450b06/godoc/vfs/zipfs/zipfs.go#L115. That is surprising.

shurcooL:
That is unexpected indeed.

I've made https://go-review.googlesource.com/#/c/19503 to fix the issue upstream.

shurcooL:
That CL has been submitted (accepted) by now, so we can revert this change. Of course, we'd need to update to the latest `x/tools` revision wherever that package is vendored.

PR #251, opened 2016-02-24
validate_cmd: not checking this error is causing a nil dereference.

jpoler:
@beyang

### DIFF
--- a/cli/validate_cmd.go
+++ b/cli/validate_cmd.go
@@ -113,7 +113,10 @@ walkLoop:
 
 		pth := walker.Path()
 
-		if fi, err := os.Stat(pth); err != nil || !fi.IsDir() {
+		fi, err := os.Stat(pth)
+		if err != nil {
+			return err
+		} else if !fi.IsDir() {
 			continue
 		}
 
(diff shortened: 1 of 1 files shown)


cli/validate_cmd.go
```diff
@@ -113,8 +113,7 @@ walkLoop:
 
 		pth := walker.Path()
 
-		fi, err := os.Stat(pth)
-		if !fi.IsDir() {
+		if fi, err := os.Stat(pth); err != nil || !fi.IsDir() {
```
beyang: can we use `os.IsNotExist(err)` and if the error is not nil AND not a file-not-found error, the entire command should fail, because that is unexpected, right?

jpoler:
@beyang, nice catch. After thinking about it I think that any error at all would be unexpected, whether it is the file missing or some other error.

cli/validate_cmd.go
shurcooL: We wouldn't want to check [`os.IsPermission(err)`](https://godoc.org/os#IsPermission) or anything like that, would we?

PR #254, opened 2016-02-28
Coverage cmd

beyang:
Coverage command for srclib data

### DIFF
--- a/cli/coverage_cmd.go
+++ b/cli/coverage_cmd.go
@@ -1,0 +1,177 @@
+package cli
+
+import (
+	"bytes"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"path/filepath"
+
+	"sourcegraph.com/sourcegraph/go-flags"
+
+	"sourcegraph.com/sourcegraph/srclib/config"
+	"sourcegraph.com/sourcegraph/srclib/graph"
+	"sourcegraph.com/sourcegraph/srclib/grapher"
+	"sourcegraph.com/sourcegraph/srclib/plan"
+)
+
+func init() {
+	cliInit = append(cliInit, func(cli *flags.Command) {
+		_, err := cli.AddCommand("coverage",
+			"srclib coverage",
+			"compute approximate amount of code successfully analyzed by srclib",
+			&coverageCmd,
+		)
+		if err != nil {
+			log.Fatal(err)
+		}
+	})
+}
+
+type Coverage struct {
+	FileScore      float32  // % files successfully processed
+	RefScore       float32  // % internal refs that resolve to a def
+	TokDensity     float32  // average number of refs/defs per LoC
+	UncoveredFiles []string // files for which srclib data was not successfully generated (best-effort guess)
+}
+
+type codeFileDatum struct {
+	LoC          int
+	NumRefs      int
+	NumDefs      int
+	NumRefsValid int
+}
+
+type CoverageCmd struct {
+}
+
+var coverageCmd CoverageCmd
+
+func (c *CoverageCmd) Execute(args []string) error {
+	repo, err := OpenLocalRepo()
+	if err != nil {
+		return err
+	}
+
+	cvg, err := coverage(repo)
+	if err != nil {
+		return err
+	}
+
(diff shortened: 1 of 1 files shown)


cli/coverage_cmd.go
```diff
@@ -0,0 +1,186 @@
...
+	out, err := json.MarshalIndent(cvg, "", "  ")
+	if err != nil {
+		return err
+	}
+	fmt.Println(string(out))
+
+	return nil
+}
+
+var codeExts = []string{".go", ".java", ".py", ".rb", ".cpp", ".ts", ".cs"} // codeExt lists all file extensions that indicate a code file we want to cover
```
jpoler: Is `.java` the only extension for java code? (I'm asking out of Ignorance.)

beyang: Yes.



    _____________________________

cli/coverage_cmd.go
```diff
@@ -0,0 +1,186 @@
...
+	filepath.Walk(repo.RootDir, func(path string, info os.FileInfo, err error) error {
+		if filepath.IsAbs(path) {
+			var err error
+			path, err = filepath.Rel(repo.RootDir, path)
+			if err != nil {
+				return err
+			}
+		}
+
+		if info.IsDir() {
```
jpoler: It may also be worthwhile to check whether the directory name starts with `.` and if so return `filepath.SkipDir`.

cli/coverage_cmd.go
```diff
@@ -0,0 +1,186 @@
...
+				datum.NumDefs++
+			}
+		}
+	}
+
+	var fileTokThresh float32 = 0.7
+	numIndexedFiles := 0
+	numDefs, numRefs, numRefsValid := 0, 0, 0
+	loc := 0
+	var uncoveredFiles []string
```
jpoler: mega nit: `nonCoveredFiles` might be a little bit more clear.

cli/coverage_cmd.go
```diff
@@ -0,0 +1,186 @@
...
+		defKeys := make(map[graph.DefKey]struct{})
+		for _, def := range data.Defs {
+			defKeys[def.DefKey] = struct{}{}
+		}
+		var validRefs []*graph.Ref
+		for _, ref := range data.Refs {
+			if datum, exists := codeFileData[ref.File]; exists {
+				datum.NumRefs++
+
+				if ref.DefRepo != "" {
```
jpoler: In the long run if we want to we could query `The Graph Itself` to find out whether the ref is to a valid repo and def. That would be an over-optimization for now though.

cli/coverage_cmd.go
```diff
@@ -0,0 +1,186 @@
...
+		}
+
+		for _, def := range data.Defs {
+			if datum, exists := codeFileData[def.File]; exists {
+				datum.NumDefs++
+			}
+		}
+	}
+
+	var fileTokThresh float32 = 0.7
```
jpoler: Can we potentially do better than this? I'm not sure what a reasonable expectation is so I will trust your judgement here.

cli/coverage_cmd.go
beyang: Why? Don't we want to descend into the dir?

cli/coverage_cmd.go
beyang: Agree, especially since the dependent may not yet be processed.

cli/coverage_cmd.go
beyang: I think 0.7 is reasonable, since many lines are comment lines and thus have no tokens.

cli/coverage_cmd.go
jpoler: Well, the type of directory I had in mind specifically would be `.git`

jpoler:
LGTM. This is a vast improvement over my original hack.

cli/coverage_cmd.go
beyang: Why? "Noncovered" is not a word, but "uncovered" means "not covered," which is what we mean here.

cli/coverage_cmd.go
beyang: Oh, I see. Good point.

PR #256, opened 2016-03-12
cli: properly handle reading an empty json file

emil2k:
This is related to srclib import failing with io.EOF errors while
reading a JSON file. An io.EOF file is thrown by the JSON decoder if it
is passed a io.Reader that immediately sends an io.EOF ( i.e. an empty
file ).

@beyang

### DIFF
--- a/cli/util.go
+++ b/cli/util.go
@@ -2,6 +2,7 @@ package cli
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"log"
 	"math"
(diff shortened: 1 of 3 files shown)


cli/util.go
```diff
@@ -49,7 +50,16 @@ func PrintJSON(v interface{}, prefix string) {
 	colorable.Println(string(data))
 }
 
+var errEmptyJSONFile = errors.New("empty json file")
```
beyang: s/json/JSON/

beyang:
why are there empty JSON files in the first place?

emil2k:
Not sure, but this will improve our logs and handle the case properly. Let's deploy this to confirm that this is the issue and then we can look into why we are getting empty JSON files.

beyang:
K, LGTM deploy it, can you file the follow-up issue on yourself?

emil2k:
Can you merge this? I don't have access to merge it in.

PR #259, opened 2016-03-14
Fix urls of toolchains: s/sourcegraph.com/github.com/

artagnon:
The sourcegraph.com urls 404. Once that's fixed, everything seems fine.


beyang:
Thanks for the contributions!

PR #260, opened 2016-03-16
docs/install: elaborate on usage (store subcommands) + add zsh completion

artagnon:
.srclib-cache is fine, but we really need .srclib-store, which can be generated
using:

  $ srclib store import

Then, let's prompt the user to execute her first command that actually does
something with the data:

  $ srclib store defs

Also, add zsh completion.


beyang:
Awesome, thanks!

PR #261, opened 2016-03-16
fixing srclib sourcegraph link on homepage

samlavery:
current link on srclib.org points to a 404 on sourcegraph.com

### DIFF
--- a/docs/theme/home.html
+++ b/docs/theme/home.html
@@ -27,7 +27,7 @@
         href="#install"><button type="button" class="btn
         btn-default"><i class="fa fa-download"></i> Try it now</button></a>
         <a class="button-link" target="_blank"
-           href="https://sourcegraph.com/sourcegraph/srclib">
+           href="https://sourcegraph.com/github.com/sourcegraph/srclib">
           <button type="button" class="btn btn-default">&#x2731; Sourcegraph</button>
         </a>
         <a class="button-link" target="_blank" href="https://github.com/sourcegraph/srclib"><button type="button" class="btn btn-default"><i class="fa fa-github"></i> GitHub</button></a>


docs/theme/home.html
```diff
@@ -16,7 +16,7 @@
         Python, JavaScript, and Ruby) with a common output format, and developer
         tools that consume this format.
 
-        <p>srclib originated inside <a target="_blank" href="https://sourcegraph.com">Sourcegraph</a>, where it powers intelligent code search over hundreds of thousands of projects.
+        <p>srclib originated inside <a target="_blank" href="https://www.sourcegraph.com">Sourcegraph</a>, where it powers intelligent code search over hundreds of thousands of projects.
```
beyang: why `www`?

samlavery: standard naming convention, serving html without the www seems wrong

samlavery: also saves us a 302 redirect

beyang: It doesn't redirect right now, though. I don't have strong opinions on the `www` vs no-`www` question, but I think we should be consistent. Currently, all other URLs omit the `www`, so let's do the same here. #bikeshedding

samlavery: fixed, but at some point we will probably want to move them all to www, simply to manage cookie leaks and scale out other non-www services, but that's a problem for another day.

docs/theme/home.html
```diff
@@ -27,7 +27,7 @@
         href="#install"><button type="button" class="btn
         btn-default"><i class="fa fa-download"></i> Try it now</button></a>
         <a class="button-link" target="_blank"
-           href="https://sourcegraph.com/sourcegraph/srclib">
+           href="https://www.sourcegraph.com/github.com/sourcegraph/srclib">
```
beyang: still www?

docs/theme/home.html
```diff
@@ -27,7 +27,7 @@
         href="#install"><button type="button" class="btn
         btn-default"><i class="fa fa-download"></i> Try it now</button></a>
         <a class="button-link" target="_blank"
-           href="https://sourcegraph.com/sourcegraph/srclib">
+           href="https://ourcegraph.com/github.com/sourcegraph/srclib">
```
shurcooL: `ourcegraph.com` is incorrect, it should be `sourcegraph.com`.

PR #263, opened 2016-03-24
graph: graph all GoPackage units together, for efficiency

pararthshah:
Depends on https://github.com/sourcegraph/srclib-go/pull/78

This uses the updated srclib-go grapher API to graph all GoPackage
units in the repo in a single call to srclib-go. This avoids the
duplicate work done to load and parse dependencies of the Go packages.

The GraphMultiUnitsRule rule will output a single GoPackage.graph.json
for the entire repo, with all defs, refs and docs in a single json file.

Observed 10x reduction in total CPU time used for building sourcegraph/sourcegraph
repo in local testing.

TODO: modify cli.Import to handle GraphMultiUnitsRule. This might be simplified
with the graph store redesign.

### DIFF
--- a/cli/internal_data_cmds.go
+++ b/cli/internal_data_cmds.go
@@ -4,16 +4,24 @@ import (
 	"encoding/json"
 	"log"
 	"os"
+	"path/filepath"
 
 	"sourcegraph.com/sourcegraph/go-flags"
 
 	"sourcegraph.com/sourcegraph/srclib/graph"
 	"sourcegraph.com/sourcegraph/srclib/grapher"
+	"sourcegraph.com/sourcegraph/srclib/plan"
+	"sourcegraph.com/sourcegraph/srclib/unit"
 )
 
 func init() {
 	cliInit = append(cliInit, func(cli *flags.Command) {
 		c, err := cli.AddCommand("internal", "(internal subcommands - do not use)", "Internal subcommands. Do not use.", &struct{}{})
+		if err != nil {
+			log.Fatal(err)
+		}
+
+		_, err = c.AddCommand("emit-unit-data", "", "", &emitUnitDataCmd)
 		if err != nil {
 			log.Fatal(err)
 		}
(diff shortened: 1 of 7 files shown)


grapher/rule.go
```diff
@@ -17,6 +18,7 @@ import (
 )
 
 const graphOp = "graph"
+const goUnitType = "GoPackage"
```
beyang: Is there any way we can avoid special casing here? Ideally, srclib core should not contain any language-specific code. It's very easy for this to become a tangled mess of special cases / if statements.

beyang: Was there no way to pass existing `.a` files to the Go compiler in srclib-go?

beyang: Maybe this is something that a toolchain passes to the Register function? (whether or not it wants to receive all source units at once in the graph phase or to have separate graph phases for each source unit?)

cli/internal_data_cmds.go
```diff
@@ -44,13 +84,72 @@ func (c *NormalizeGraphDataCmd) Execute(args []string) error {
...
+		}
+
+		if _, err := os.Stdout.Write(data); err != nil {
+			return err
+		}
 	}
 
-	if _, err := os.Stdout.Write(data); err != nil {
-		return err
+	graphPerUnit := make(map[string]*graph.Output)
```
beyang: Add a comment describing what this block of code does.

cli/store_cmds.go
```diff
@@ -244,6 +244,57 @@ func Import(buildDataFS vfs.FileSystem, stor interface{}, opt ImportOpt) error {
...
+				return nil
+			}
+			if os.IsNotExist(err) {
+				log.Printf("Warning: no build data for unit %s %s.", sourceUnit.Type, sourceUnit.Name)
+				return nil
+			}
+			return fmt.Errorf("error reading JSON file %s for unit %s %s: %s", graphFile, sourceUnit.Type, sourceUnit.Name, err)
+		}
+		if opt.DryRun || GlobalOpt.Verbose {
+			log.Printf("# Importing graph data (%d defs, %d refs, %d docs, %d anns) for unit %s %s", len(data.Defs), len(data.Refs), len(data.Docs), len(data.Anns), sourceUnit.Type, sourceUnit.Name)
```
beyang: remove debug statement?

cli/store_cmds.go
```diff
@@ -244,6 +244,57 @@ func Import(buildDataFS vfs.FileSystem, stor interface{}, opt ImportOpt) error {
...
+			}
+		case store.MultiRepoImporter:
+			if err := imp.Import(opt.Repo, opt.CommitID, sourceUnit, data); err != nil {
+				return fmt.Errorf("error running store.MultiRepoImporter.Import: %s", err)
+			}
+		default:
+			return fmt.Errorf("store (type %T) does not implement importing", stor)
+		}
+
+		mu.Lock()
```
beyang: Why does this mutex exist at all? The assignment is atomic and the mutex is not consulted when the value is read. Shouldn't we just delete the mutex and add a comment to the var saying where it could be modified/read concurrently?

beyang:
lgtm after comments addressed.

cli/store_cmds.go
pararthshah: This is not for debugging, but for dry runs and 'verbose' output during builds, when the `-v` argument is passed to the command.

cli/store_cmds.go
pararthshah: Good catch, i've removed it.

PR #264, opened 2016-03-28
`srclib coverage` improvements

alexsaveliev:
- when computing density, excluding blank and commented out lines (assuming that comment starts with //)
- when processing repository with multiple units, collecting all definitions before looking for them in the references. Otherwise we may found no definition if it was defined in a different unit
- using {unit, unit type, path} when matching refs to defs; normalizing ref.DefKey and def.DefKey objects before matching them. This is done to prevent cache misses when ref points to definition from the other unit (def has no unit/unittype set while ref includes not-empty defunit/defunitype)
- Windows support (using slashes as separators)

### DIFF
--- a/cli/coverage_cmd.go
+++ b/cli/coverage_cmd.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 
 	"strings"
+	"unicode"
 
 	"sourcegraph.com/sourcegraph/go-flags"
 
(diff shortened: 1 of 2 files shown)


cli/coverage_cmd.go
```diff
@@ -196,3 +212,63 @@ func divideSentinel(x, y, sentinel float64) float64 {
...
+
+	len := len(data)
+	if len == 0 {
+		return 0
+	}
+
+	count := 1
+	start := 0
+
+	comment := []byte{'/', '/'}
```
shurcooL: Optional style suggestion, this might be nicer/more readable to write using this form:

```Go
comment := []byte("//")
```

End result is identical.

Also, if you're calling this `numLines` func a lot, maybe it's worth it to move `comment` declaration out so it's not allocated each time this runs. (But maybe it's not worth it.)

PR #266, opened 2016-03-29
`srclib coverage` improvements

alexsaveliev:
- excluding comments when counting lines of code, otherwise small thoroughly commented file (usually with the large copyright or license information) may be considered 'uncovered'

### DIFF
--- a/cli/coverage_test.go
+++ b/cli/coverage_test.go
@@ -1,8 +1,30 @@
 package cli
 
 import (
+	"io/ioutil"
 	"testing"
 )
+
+func TestStripComments(t *testing.T) {
+
+	source, _ := ioutil.ReadFile("testdata/strip-comments/commented-source.txt")
+	target, _ := ioutil.ReadFile("testdata/strip-comments/expected-source.txt")
+	expected := string(target)
+
+	actual := string(stripComments(source))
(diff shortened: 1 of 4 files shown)


cli/coverage_test.go
```diff
@@ -1,9 +1,31 @@
...
 
 import (
+	"io/ioutil"
 	"testing"
 )
 
+func TestStripComments(t *testing.T) {
+
+	source, _ := ioutil.ReadFile("testdata/strip-comments/commented-source.txt")
+	target, _ := ioutil.ReadFile("testdata/strip-comments/expected-source.txt")
```
beyang: You can just inline the text instead of storing them as separate files. Just use the Go string literal syntax:

```
str := `this
is a
multi line 
string literal`

```

beyang:
One comment, otherwise LGTM

cli/coverage_test.go
alexsaveliev: I'd prefer to leave large chunks of text in a separate file, otherwise code becomes unreadable because there would be a mix of Go code and string data that looks quite similar to Go code.
My understanding is that `testdata` is a convention where test resources should be kept, content of this directory is ignored by `govendor` and `godep`

Issue #267, opened 2016-03-30
Failed to install Go toolchain


alexsaveliev:
I think that pre-requisite is to have empty GOPATH, for example with `export GOPATH=/tmp/nogo`
```
user@user-virtual-machine:~$ srclib toolchain install go
Go (github.com/sourcegraph/srclib-go) =========================================
mkdir -p /tmp/nogo/src/github.com/sourcegraph
ln -s /home/user/.srclib/github.com/sourcegraph/srclib-go /tmp/nogo/src/github.com/sourcegraph/srclib-go
Symlinked toolchain github.com/sourcegraph/srclib-go into your GOPATH at /tmp/nogo/src/github.com/sourcegraph/srclib-go
Downloading Go toolchain
Cloning into 'srclib-go'...
remote: Counting objects: 5077, done.
remote: Compressing objects: 100% (159/159), done.
remote: Total 5077 (delta 56), reused 0 (delta 0), pack-reused 4883
Receiving objects: 100% (5077/5077), 2.87 MiB | 1.27 MiB/s, done.
Resolving deltas: 100% (1708/1708), done.
Checking connectivity... done.
Building Go toolchain program
Running [make] in /home/user/.srclib/github.com/sourcegraph/srclib-go
GOBIN=/home/user/.srclib/github.com/sourcegraph/srclib-go/.bin go get github.com/tools/godep
.bin/godep go build -o .bin/srclib-go
godep: [WARNING]: godep should only be used inside a valid go package directory and
godep: [WARNING]: may not function correctly. You are probably outside of your $GOPATH.
godep: [WARNING]:       Current Directory: /home/user/.srclib/github.com/sourcegraph/srclib-go
godep: [WARNING]:       $GOPATH: /tmp/nogo
godep: WARNING: Go version (go1.6) & $GO15VENDOREXPERIMENT= wants to enable the vendor experiment, but disabling because a Godep workspace (Godeps/_workspace) exists
graph.go:18:2: cannot find package "sourcegraph.com/sourcegraph/srclib-go/gog" in any of:
        /usr/local/go/src/sourcegraph.com/sourcegraph/srclib-go/gog (from $GOROOT)
        /home/user/.srclib/github.com/sourcegraph/srclib-go/Godeps/_workspace/src/sourcegraph.com/sourcegraph/srclib-go/gog (from $GOPATH)
        /tmp/nogo/src/sourcegraph.com/sourcegraph/srclib-go/gog
[...]

PR #268, opened 2016-03-30
Fixed installation of Go-based toolchains

alexsaveliev:
- `govendor` (which is currently used by srclib-go and srclib-python) refuses to synchronize when runs not from GOPATH.
To prevent build failures, running `make` from GOPATH/src/TOOLCHAIN instead of SRCLIBDIR/TOOLCHAIN
- fixes sourcegraph/srclib#267


beyang:
Not all toolchains will be installed under `$GOPATH`. What about modifying `$GOPATH` so that it includes `$SRCLIBDIR`?

PR #271, opened 2016-04-01
Introduced SkipScan Srcfile's flag

alexsaveliev:
- introducing new flag SkipScan of Srcfile that may disable 'scan' phase execution (assumed that Srcfile contains all the source units and there is no need to scan for extra)


alexsaveliev:
As suggested by Beyang, I can use `Scanners: []` instead

PR #272, opened 2016-04-04
Change srclib metadata import to write new refs/defs/indexes in serial.

samlavery:
This change removes the parallel writes that were in place previously.
This is only a change at the source unit level, multiple source units can still be stored in parallel, but this
change makes the writes atomic within a single source unit.  In theory the old parallel writes should have made import faster, but on slower storage systems, this was causing write contention and data
corruption, slowing things way down.  Serial writes eliminate this corruption.

### DIFF
--- a/store/indexed.go
+++ b/store/indexed.go
@@ -596,22 +596,20 @@ func (s *indexedUnitStore) Import(data graph.Output) error {
 	var defOfs, refOfs byteOffsets
 	var refFBRs fileByteRanges
 
-	par := parallel.NewRun(2)
-	par.Do(func() (err error) {
-		defOfs, err = s.fsUnitStore.writeDefs(data.Defs)
-		return err
-	})
-	par.Do(func() (err error) {
-		refFBRs, refOfs, err = s.fsUnitStore.writeRefs(data.Refs)
-		return err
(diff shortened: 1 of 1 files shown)


store/indexed.go
```diff
@@ -596,22 +596,20 @@ func (s *indexedUnitStore) Import(data graph.Output) error {
...
 
-	par := parallel.NewRun(2)
-	par.Do(func() (err error) {
-		defOfs, err = s.fsUnitStore.writeDefs(data.Defs)
-		return err
-	})
-	par.Do(func() (err error) {
-		refFBRs, refOfs, err = s.fsUnitStore.writeRefs(data.Refs)
+	//Complete each write operation serially which eliminates
+	//a lock contention bug on slower underlying file stores
```
shurcooL: Style nitpick, Go comments should have a space between `//` and the content. Not having a space is used for executable directives, e.g., `//go:generate` and temporarily commented out code (which usually doesn't get committed), e.g., `//log.Println(someDebugValue)`.

You can see good examples of comments in [this file](https://sourcegraph.com/github.com/golang/go@master/-/tree/src/net/http/transfer.go), for example.

PR #275, opened 2016-04-06
Coverage by lang

beyang:
Calculate coverage per language for greater accuracy.

### DIFF
--- a/cli/coverage_cmd.go
+++ b/cli/coverage_cmd.go
@@ -25,6 +25,8 @@ import (
 	"sourcegraph.com/sourcegraph/srclib/unit"
 )
 
+const fileTokThresh float64 = 0.7
+
 func init() {
 	cliInit = append(cliInit, func(cli *flags.Command) {
 		_, err := cli.AddCommand("coverage",
@@ -43,6 +45,7 @@ type codeFileDatum struct {
 	NumRefs      int
 	NumDefs      int
 	NumRefsValid int
+	Language     string
 }
 
 type CoverageCmd struct {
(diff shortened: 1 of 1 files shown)


cli/coverage_cmd.go
```diff
@@ -70,18 +73,32 @@ func (c *CoverageCmd) Execute(args []string) error {
...
+	"Python":      []string{".py"},
+	"Ruby":        []string{".rb"},
+	"C++":         []string{".cpp"},
+	"TypeScript":  []string{".ts"},
+	"C#":          []string{".cs"},
+	"JavaScript":  []string{".js"},
+	"PHP":         []string{".php"},
+	"Objective-C": []string{".m"},
+}
+var extToLang map[string]string
```
alexsaveliev: What if there are multiple file extensions denoting the same language (example: .c++, .cxx, .cpp, .cc)?

cli/coverage_cmd.go
```diff
@@ -100,22 +117,23 @@ func coverage(repo *Repo) (*cvg.Coverage, error) {
 
 		path = filepath.ToSlash(path)
 
-		if _, isCodeFile := codeExts_[filepath.Ext(path)]; isCodeFile {
+		ext := filepath.Ext(path)
```
alexsaveliev: I'd suggest to normalize extensions by making them lowercase

cli/coverage_cmd.go
beyang: I'll add `.cc` for C++

cli/coverage_cmd.go
```diff
@@ -144,11 +162,11 @@ func coverage(repo *Repo) (*cvg.Coverage, error) {
 		data = append(data, item)
 
 		for _, def := range item.Defs {
-			defKeys[adjustDefKey(def.DefKey, sourceUnit)] = struct{}{}
+			defKeys[def.DefKey] = struct{}{}
 		}
 
 		for _, ref := range item.Refs {
-			ref.SetFromDefKey(adjustDefKey(ref.DefKey(), sourceUnit))
+			ref.SetFromDefKey(ref.DefKey())
```
alexsaveliev: Not needed if we about to drop `adjustDefKey`, the purpose of `ref.SetFromDefKey` was to update ref's data

cli/coverage_cmd.go
```diff
@@ -192,29 +209,49 @@ func coverage(repo *Repo) (*cvg.Coverage, error) {
...
+	stats := make(map[string]*langStats)
 	for file, datum := range codeFileData {
-		loc += datum.LoC
-		numDefs += datum.NumDefs
-		numRefs += datum.NumRefs
-		numRefsValid += datum.NumRefsValid
+		if _, exist := stats[datum.Language]; !exist {
+			stats[datum.Language] = &langStats{}
+		}
+		if filepath.Base(file) == "doc.go" {
```
alexsaveliev: I'd refactor this block and extract a function that determines if file should be taken into account

cli/coverage_cmd.go
alexsaveliev: - `.mm` for Objective-C
- `.cxx` for C++
- `.c++` for C++
- `.h` for ...?

beyang: punt on ".h" for now

sqs: Can we use the existing filelang pkg to determine this? Should we?

PR #276, opened 2016-04-06
Updated sqs/fileset

alexsaveliev:
- Updated sqs/fileset to include fix for sqs/fileset#2

### DIFF
--- a/vendor/vendor.json
+++ b/vendor/vendor.json
@@ -163,10 +163,10 @@
 			"revisionTime": "2014-11-13T16:00:59-07:00"
 		},
 		{
-			"checksumSHA1": "99uvKlT3u4DB243YwHceIMOBZKg=",
+			"checksumSHA1": "uxEPICJVR/rl0zLPyeFuJh8CAuM=",
 			"path": "github.com/sqs/fileset",
-			"revision": "af79d58d268e6fb30711babccd37156c95fd5aaf",
-			"revisionTime": "2016-01-27T15:26:27-08:00"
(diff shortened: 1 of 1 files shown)


shurcooL:
Updating sqs/fileset dependency to sqs/fileset@012a6d31cbc4d10f3148918e0a37cd333db58a34, which is its latest `master` version, LGTM.

But I am very puzzled by the `.gitignore` of this repo ignoring all of /vendor/` contents except `vendor.json`:

https://github.com/sourcegraph/srclib/blob/488ce0095f1a04e713bed8c715017b28796b3610/.gitignore#L4

How does that work, and what's the motivation to do vendoring while not actually committing anything to the repository?

alexsaveliev:
> How does that work, and what's the motivation to do vendoring while not actually committing anything to the repository?

Default make target does `govendor sync` which pulls all the dependencies. 
I think having both `vendor.json` descriptor and source code of dependencies is incorrect. The same way NodeJS-based source code is organized - package.json checked in into SCM, dependencies are fetched on demand.
If you think we should keep dependencies source code - please let me know and I'll add it back. See also similar PR sourcegraph/srclib-go#80. Please note that srclib-go had an example why it's bad to keep both descriptor and source code - Godeps referred to the non-existent hash while dependency source code was there. Looks like somebody forgot to update either hash or dependency source code

shurcooL:
> If you think we should keep dependencies source code - please let me know and I'll add it back.

I don't have strong opinions on how vendoring is done, so no need to make immediate changes. It was a question.

I didn't know about `govendor sync` command, but I see how it works now, thanks for explaining.

PR #277, opened 2016-04-08
Fix concurrency issues in import step

pararthshah:
This PR fixes two issues that were recently introduced with the `GraphMultiUnitsRule`:

1. Protect the shared variable `hasIndexableData` with a mutex to avoid Go race detector from complaining.

2. Use proper closure for concurrent to `importGraphData` calls within for loop. This issue would cause the graph data for certain units from never getting imported, while certain other units would be imported multiple times, within the same build.

### DIFF
--- a/cli/store_cmds.go
+++ b/cli/store_cmds.go
@@ -246,7 +246,13 @@ func Import(buildDataFS vfs.FileSystem, stor interface{}, opt ImportOpt) error {
 	// need to be protected by a mutex since its value is only modified in one
 	// direction (false to true), and it is only read in the sequential section
 	// after parallel.NewRun completes.
-	var hasIndexableData bool
+	//
+	// However, we still protect it with a mutex to avoid data race errors from the
+	// Go race detector.
+	var (
+		mu               sync.Mutex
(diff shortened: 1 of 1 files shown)


beyang:
LGTM

cli/store_cmds.go
```diff
@@ -309,7 +317,8 @@ func Import(buildDataFS vfs.FileSystem, stor interface{}, opt ImportOpt) error {
 				return importGraphData(rule.Target(), rule.Unit)
 			})
 		case *grapher.GraphMultiUnitsRule:
-			for target, sourceUnit := range rule.Targets() {
+			for target_, sourceUnit_ := range rule.Targets() {
+				target, sourceUnit := target_, sourceUnit_
```
shurcooL: FWIW, style-wise, I think it would've been fine not to add underscores. E.g.:

```Go
target, sourceUnit := target, sourceUnit
```

pararthshah: I see, I followed the `rule := rule_` style from the enclosing loop; I don't feel strongly about either way.

PR #278, opened 2016-04-08
Restored Windows support for Go toolchain

alexsaveliev:
- `os.Symlink` makes "file-type" symbolic links on Windows while `cd foo && make` requires "foo" to be either directory or so-called "directory symbolic link". This was the reason that  it was impossible to install go toolchain on Windows because attempt to execute `make` in file-type symbolic link failed. I also had to change execution order from (symlink, clone) to (clone, symlink) when installing Go toolchain
- thou shall not emit backslashes when generating Makefile commands, they being cut making command not runnable

### DIFF
--- a/cli/toolchain_cmd.go
+++ b/cli/toolchain_cmd.go
@@ -9,6 +9,7 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
+	"runtime"
 
 	"strings"
 
@@ -267,7 +268,7 @@ var stdToolchains = toolchainMap{
 
 func (m toolchainMap) listKeys() string {
 	var langs string
-	for i, _ := range m {
+	for i := range m {
 		langs += i + ", "
 	}
 	// Remove the last comma from langs before returning it.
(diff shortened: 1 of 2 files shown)


cli/toolchain_cmd.go
```diff
@@ -612,9 +614,15 @@ func symlinkToGopath(toolchain string) (err error, gopathDir string) {
...
-		log.Printf("ln -s %s %s", srclibpathDir, gopathDir)
-		if err := os.Symlink(srclibpathDir, gopathDir); err != nil {
-			return err, ""
+		if runtime.GOOS != "windows" {
+			log.Printf("ln -s %s %s", srclibpathDir, gopathDir)
+			if err := os.Symlink(srclibpathDir, gopathDir); err != nil {
+				return err, ""
+			}
+		} else {
+			if err := execCmdInDir(srclibpathDir, "cmd", "/c", "mklink", "/D", gopathDir, srclibpathDir); err != nil {
```
beyang: Add a comment explaining why os.Symlink doesn't work

beyang:
After comment is addressed, LGTM

PR #285, opened 2016-04-27
Coverage stats improvements

alexsaveliev:
- when calculating number of lines of code, stripping everything except idents. I think that line of code that contains no idents probably cannot be indexed - there is nothing to extract from. Examples may be: literals, comments
- separating uncovered and undiscovered files. "Undiscovered" file is a file that matches code file by extension but not listed in the discovered source units. "Uncovered" file was discovered during the scan phase but its coverage density is too small


cli/coverage_cmd.go
```diff
@@ -329,7 +345,7 @@ func stripComments(data []byte) []byte {
 	tok := s.Scan()
 	for tok != scanner.EOF {
 		pos := s.Pos()
-		if tok == scanner.Comment {
+		if tok != scanner.Ident {
```
beyang: Is the Go scanner accurate enough for other languages? I think it is, but perhaps it's worth adding a comment for a future reader of the code (some languages' comment syntax, e.g., Python, is a bit weird).

beyang:
1 comment, otherwise LGTM1

PR #289, opened 2016-05-09
Change to srclib build planning to allow for extremely large reposito…

samlavery:
…ries and number of source units.  Fixes the k8s build error.

### DIFF
--- a/grapher/rule.go
+++ b/grapher/rule.go
@@ -4,8 +4,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"strings"
-
+	
 	"sourcegraph.com/sourcegraph/makex"
 	"sourcegraph.com/sourcegraph/srclib"
 	"sourcegraph.com/sourcegraph/srclib/buildstore"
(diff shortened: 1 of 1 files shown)


grapher/rule.go
```diff
@@ -150,7 +149,12 @@ func (r *GraphMultiUnitsRule) Recipes() []string {
...
+	// very long command line was resulting in build failures for large repos. We were seeing spawn/fork errors specifically building the kubernetes
+	// repository.  k8s has so many dependencies that the cli argument string to srclib exceeded the alotted stack size for a newly spawned process.
+	// Find and xargs allows us to emulate having all the source units without having to significantly change how we handle process to process communication.   
 	return []string{
-		fmt.Sprintf("%s internal emit-unit-data %s | %s tool %q %q | %s internal normalize-graph-data --unit-type %q --dir . --multi --data-dir %s", safeCommand, strings.Join(unitFiles, " "), safeCommand, r.Tool.Toolchain, r.Tool.Subcmd, safeCommand, r.UnitsType, filepath.ToSlash(r.dataDir)),
+		fmt.Sprintf("find .srclib-cache -name \"*%s.unit.json\" | xargs %s internal emit-unit-data  | %s tool %q %q | %s internal normalize-graph-data --unit-type %q --dir . --multi --data-dir %s", r.UnitsType, safeCommand, safeCommand, r.Tool.Toolchain, r.Tool.Subcmd, safeCommand, r.UnitsType, r.dataDir),
```
shurcooL: Suggestion, consider using a raw literal string instead of interpreted one, it should be more readable with one less layer of escaping:

```Go
fmt.Sprintf(`find .srclib-cache -name "*%s.unit.json" | xargs %s internal emit-unit-data  | %s tool %q %q | %s internal normalize-graph-data --unit-type %q --dir . --multi --data-dir %s`, r.UnitsType, safeCommand, safeCommand, ...)
```

beyang: Why did we get rid of the `filepath.ToSlash` wrapper around r.dataDir?

beyang:
merged separately

grapher/rule.go
keegancsmith: Also is the output of find the same as what would be in `unitFiles`?

alexsaveliev: I'd suggest to use `/usr/bin/find`, at least when `runtime.GOOS == "windows"` otherwise wrong `find` may be used

beyang: done.

beyang: @keegancsmith in this case it is. It's a bit of a hack, but `GraphMultiUnitsRule` as a whole is a bit of a hack and hopefully will be rendered unnecessary by incremental compile.

Issue #250, opened 2016-02-19
Download of Go toolchain fails because of redirects


bkabrda:
Downloading Go toolchain has starting failing for me recently:
```
Downloading Go toolchain
Cloning into 'srclib-go'...
fatal: repository 'https://sourcegraph.com/sourcegraph/srclib-go/' not found
failed to install/upgrade Go (sourcegraph.com/sourcegraph/srclib-go) toolchain: exit status 128
```
The problem is that git client can't handle redirects, but the above url redirects:
```
$ curl -I sourcegraph.com/sourcegraph/srclib-go
HTTP/1.1 301 Moved Permanently
```
The redirect target is https://sourcegraph.com/github.com/sourcegraph/srclib.
Everything was working several days ago, so maybe you just setup the redirects recently? Or am I doing something wrong?
Thanks

shurcooL:
Thanks for reporting this. It's a configuration change we've done our side recently that caused this, we'll work on fixing it. /cc @pararthshah

The redirect target should be https://sourcegraph.com/github.com/sourcegraph/srclib-go and that would work as a `git clone` target for now.

guilhermebr:
:+1:

pararthshah:
This is fixed by commit [f37ecd8](https://github.com/sourcegraph/srclib/commit/f37ecd88157b8a4606035ca28ceddc2a09d4ef99), as now all toolchains are fetched directly from GitHub.

PR #292, opened 2016-05-20
Downloading toolchains from GitHub

alexsaveliev:
- using github.com/sourcegraph/toolchain to clone toolchains from (sourcegraph.com/sourcegraph/toolchain is deprecated) but placing toolchains in sourcegraph.com/sourcegraph/toolchain
- installing toolchains during Travis build


cli/toolchain_cmd.go
```diff
@@ -663,3 +669,20 @@ func symlink(source, target string) error {
...
 }
+
+// prepareParentDir makes parent directory for a given toolchain under SRCLIBPATH and returns SRCLIBPATH/FULL-PATH-TO-TOOLCHAIN
+func prepareParentDir(toolchainName string) (toolchainDir string, err error) {
+	// toolchain dir under SRCLIBPATH
+	srclibpathDir := filepath.Join(filepath.SplitList(srclib.Path)[0], toolchainFqn(toolchainName))
+	return srclibpathDir, os.MkdirAll(filepath.Dir(srclibpathDir), 0700)
+}
+
+// toolchainFqn returns fully qualified toolchain name
```
shurcooL: Go style says that acronyms should be all upper case, see https://github.com/golang/go/wiki/CodeReviewComments#initialisms.

cli/toolchain_cmd.go
```diff
@@ -663,3 +669,20 @@ func symlink(source, target string) error {
...
+func toolchainFqn(toolchainName string) string {
+	return "sourcegraph.com/sourcegraph/" + toolchainName
+}
+
+// toolchainFqn returns clone URI for a given toolchain
+func toolchainCloneUri(toolchainName string) string {
```
shurcooL: Same as above, this should be `toolchainCloneURI` not `toolchainCloneUri`.

cli/toolchain_cmd.go
```diff
@@ -605,7 +611,7 @@ func cloneToolchain(dest, toolchain string) error {
 			return err
 		}
 
-		cmd := exec.Command("git", "clone", "https://"+toolchain)
+		cmd := exec.Command("git", "clone", toolchain)
```
shurcooL: Please rename `toolchain` to `cloneURL` or `cloneURI`. The func is already called `cloneToolchain`, so it's clear that we're dealing with toolchains, but `toolchain` variable is very hard to understand what it actually is.

.travis.yml
```diff
@@ -1,19 +1,36 @@
 language: go
 go: 1.6
-sudo: false
+sudo: required
```
shurcooL: Is it tricky to get this to keep running in Travis' container-based infrastructure? Just asking, no need to change this if it's really hard or time consuming.

shurcooL:
Overall, looks great. I left some comments.

.travis.yml
alexsaveliev: > `sudo` is not required if we want to run only Go tests. However, if we want to make sure all our toolchains can be cloned and installed (in the last few months we had a few similar issues (#267, #250,  sourcegraph/srclib-javascript#31), we need to install packages and applications using `apt-get`, which requires `sudo` permissions. Maybe some day I'll try to download and install packages for current user only without using package manager but that's another story :)

alexsaveliev:
@shurcooL thanks for your feedback, addressed your comments

cli/toolchain_cmd.go
```diff
@@ -604,24 +604,24 @@ Refusing to install Basic toolchain because %s is not installed or is not on the
...
 		cmd.Stderr = os.Stderr
 		return cmd.Run()
 	} else if err != nil {
 		return err
 	} else if !fi.Mode().IsDir() {
-		return fmt.Errorf("not a directory: %s", dest)
+		return fmt.Errorf("not a directory: %s", destDir)
 	}
-	log.Printf("Toolchain directory %q already exists, using existing version.", dest)
+	log.Printf("Toolchain directory %q already exists, using existing version.", destDir)
```
shurcooL: I wanted to see if the readability of this `cloneToolchain` func could be improved by following [Indent Error Flow style suggestion](https://github.com/golang/go/wiki/CodeReviewComments#indent-error-flow):

```Go
func cloneToolchain(destDir, cloneURI string) error {
	switch fi, err := os.Stat(destDir); {
	case err != nil && !os.IsNotExist(err):
		return err
	case err == nil && !fi.Mode().IsDir():
		return fmt.Errorf("not a directory: %s", destDir)
	case err == nil && fi.Mode().IsDir():
		log.Printf("Toolchain directory %q already exists, using existing version.", destDir)
		return nil
	}

	// Clone.
	...
}
```

It does look easier to read/reason about to me. Feel free to take this if you want. Completely optional; I just wanted to share this because the thought occurred to me.

shurcooL:
LGTM.

Of course we should test that the toolchains continue to be installable after getting this merged, but code changes are sound. 👍

PR #293, opened 2016-05-23
New source unit format

beyang:
This change simplifies the source unit data structure, removing the old `RepoSourceUnit` and `SourceUnit` and replacing it with a `SourceUnit` data structure defined in protobuf.

The motivation behind this change is to make it possible to support "global deps" (i.e., source unit resolution at the application layer) with a cleaner API.

In the process, this also deprecates / removes some configuration functionality that is unused by any existing toolchain (the `Config` field for the `Srcfile` format and the `Ops` field).

- [x] The JSON serialization of the new `SourceUnit` is backward compatible with the old `SourceUnit`.

Most of the remaining is just changes to test files in response to the data structure change.


cli/test_cmd.go
```diff
@@ -181,6 +181,7 @@ func testTree(treeDir, expectedDir, actualDir string, generateExpected bool) err
 	cmd.Stderr, cmd.Stdout = w, w
 	cmd.Env = append(os.Environ(), "SRCLIB_FOLLOW_CROSS_FS_SYMLINKS=true")
 
+	log.Printf("# %v, %+v", cmd.Dir, cmd.Args)
```
keegancsmith: do we wanna keep this debugging output?

unit/unit.go
```diff
@@ -0,0 +1,107 @@
...
+
+// UnitRepoUnresolved is a sentinel value that indicates this unit's
+// repository is unresolved
+const UnitRepoUnresolved = "?"
+
+func (u Key) IsResolved() bool {
+	return u.Repo == UnitRepoUnresolved
+}
+
+// ContainsAny returns true if u contains any files in filesnames. Currently
```
keegancsmith: nit: typo `s/filesnames/filenames/`

unit/unit.go
```diff
@@ -0,0 +1,107 @@
...
+// idSeparator joins a source unit's name and type in its ID string.
+const idSeparator = "@"
+
+// ID returns an opaque identifier for this source unit that is guaranteed to be
+// unique among all other source units in the same repository.
+func (u SourceUnit) ID() ID {
+	return ID(fmt.Sprintf("%s%s%s", url.QueryEscape(u.Name), idSeparator, u.Type))
+}
+
+func (u *SourceUnit) ID2() ID2 {
```
keegancsmith: I am missing context here. What is ID vs ID2?

unit/unit.go
```diff
@@ -0,0 +1,107 @@
...
+// ID is a source unit ID. It is only unique within a repository.
+type ID string
+
+// ID2 is a source unit ID. It is only unique within a repository.
+type ID2 struct {
+	Type string
+	Name string
+}
+
+func (v ID2) String() string { return fmt.Sprintf("{%s %s}", v.Type, v.Name) }
```
keegancsmith: why does this use a different format to what `ID()` returns?

unit/unit.go
```diff
@@ -0,0 +1,107 @@
...
+			expanded = append(expanded, filepath.ToSlash(hit))
+		}
+	}
+	return expanded, nil
+}
+
+type SourceUnits []*SourceUnit
+
+func (v SourceUnits) Len() int           { return len(v) }
+func (v SourceUnits) Less(i, j int) bool { return v[i].String() < v[j].String() }
```
keegancsmith: How much do we rely on sorting SourceUnits? If a lot, this seems like a very inefficient less function. But I assume we don't do it that much.

cli/units_cmd.go
```diff
@@ -46,22 +46,24 @@ func scanUnitsIntoConfig(cfg *config.Repository, quiet bool) error {
...
+					if u.Config == nil {
+						u.Config = map[string]string{}
+					}
+					u.Config[k] = v
 				}
-				u.Config[k] = v
 			}
 		}
-	}
+	*/
```
keegancsmith: just delete this block? Although not sure why we don't want it anymore

unit/unit.go
```diff
@@ -105,3 +107,105 @@ type SourceUnits []*SourceUnit
 func (v SourceUnits) Len() int           { return len(v) }
 func (v SourceUnits) Less(i, j int) bool { return v[i].String() < v[j].String() }
 func (v SourceUnits) Swap(i, j int)      { v[i], v[j] = v[j], v[i] }
+
+// sourceUnit is the struct for JSON serialization used to preserve
+// backcompat with the old SourceUnit JSON serialization format
+type sourceUnit struct {
```
keegancsmith: will we ever be able to remove this? How do we communicate version information on srclib data?

unit/unit.proto
```diff
@@ -24,20 +24,21 @@ message Key {
 	// CommitID is the commit ID of the repository containing this
 	// source unit, if any. The scanner tool need not fill this in; it
 	// should be left blank, to be filled in by the `srclib` tool.
-	//
-	// CommitID can also be a semantic version if the unit is unresolved.
 	string CommitID = 2;
 
+	// Version is the unresolved source unit version (e.g., "v1.2.3").
```
keegancsmith: can you expand on when this will be empty/etc like you do with the other fields?

cli/units_cmd.go
```diff
@@ -46,24 +46,24 @@ func scanUnitsIntoConfig(cfg *config.Repository, quiet bool) error {
...
+				if u.Config == nil {
+					u.Config = make(map[string]string)
+				}
+				if vstr, isStr := v.(string); isStr {
+					u.Config[k] = vstr
 				}
 			}
 		}
-	*/
+	}
```
keegancsmith: I thought this is deprecated, why bother setting it?

plan/makefile_test.go
```diff
@@ -12,20 +12,34 @@ import (
...
+				Key: unit.Key{
+					Name: "n",
+					Type: "t",
+				},
+				Info: unit.Info{
+					Files: []string{"f"},
+					Ops: map[string][]byte{
+						"graph":      nil,
+						"depresolve": nil,
+					},
```
keegancsmith: If we want to deprecate this field, should we support it being empty now?

store/test_util.go
```diff
@@ -0,0 +1,9 @@
+package store
+
+import "encoding/json"
+
+func deepEqual(u, v interface{}) bool {
```
keegancsmith: why did the old `reflect.DeepEqual` break?

keegancsmith:
LGTM!

So the most useful change to this is splitting out the `Key` part of a `SourceUnit`. That accomplishes the cleaner API goal stated in the PR description?

A lot of the commits here are fixup commits. Could you squash them before merging? Doing that before submitting for review would of also been cool :)


> In the process, this also deprecates / removes some configuration functionality that is unused by any existing toolchain (the Config field for the Srcfile format and the Ops field).

If it is unused, can't we just remove it?


> The JSON serialization of the new SourceUnit is backward compatible with the old SourceUnit.

What testing have you done for this? From reading the code it seems alright, but we have a lot of existing data you could test against.

unit/unit.go
shurcooL: Agree, and if `ID2` is going to stay, it should be documented since it's an exported identifier (but primarily to help explain why it exists).

unit/unit.go
beyang: We should be able to remove this soon after switching over. It is just needed in the interim for toolchains whose docker images are not switched over immediately to the new srclib. I will add the comment.

cli/units_cmd.go
beyang: It was necessary to pass a failing test case, which led me to believe it was used somehow, somewhere.

plan/makefile_test.go
beyang: It's still necessary until we can get rid of the multi-unit `graphAll` op hack that was introduced to optimize srclib-go perf. Richard will be removing that soon, I believe.

store/test_util.go
beyang: It does not treat empty arrays and `nil` as equal.

beyang:
squashed and pushed.

beyang:
merge-squashed

PR #295, opened 2016-05-25
gendata: Remove unneeded fmt.Sprintf.

shurcooL:
Don't use `fmt.Sprintf` here since it's not needed.

It's dangerous to provide dynamic input as the first parameter of a printf-like function since it may contain a sequence like `%v` which would cause a problem.


shurcooL:
There's a merge conflict now, but resolving it is going to be trivial - this PR is just 2 modified lines.

It just needs to be code reviewed to be able to merge.

keegancsmith:
LGTM

PR #297, opened 2016-05-26
improve scan error messages

mkovacs:
Make srclib elaborate on errors involving the scan commands to exec.

### DIFF
--- a/toolchain/find.go
+++ b/toolchain/find.go
@@ -152,16 +152,10 @@ func newInfo(toolchainPath, dir, configFile string) (*Info, error) {
 	prog := filepath.Join(".bin", filepath.Base(toolchainPath))
 	if runtime.GOOS == "windows" {
 		prog = winExe(dir, prog)
-	}
-
-	if runtime.GOOS != "windows" {
-		fi, err := os.Stat(filepath.Join(dir, prog))
-		if os.IsNotExist(err) {
-			prog = ""
-		} else {
-			if fi.Mode().Perm()&0111 == 0 {
-				return nil, fmt.Errorf("installed toolchain program %q is not executable (+x)", prog)
-			}
+	} else {
+		cmdName := filepath.Join(dir, prog)
+		if err := checkRegularExecutableFile(cmdName); err != nil {
+			return nil, err
 		}
 	}
 
@@ -171,6 +165,24 @@ func newInfo(toolchainPath, dir, configFile string) (*Info, error) {
 		ConfigFile: configFile,
 		Program:    prog,
 	}, nil
+}
+
+// checkRegularExecutableFile verifies that the file pointed to by name is a regular executable file.
+func checkRegularExecutableFile(name string) error {
+	switch info, err := os.Stat(name); {
+	case os.IsNotExist(err):
+		return fmt.Errorf("toolchain %s does not exist (should be a regular executable file)", name)
+	case err != nil:
+		return fmt.Errorf("toolchain %s caused %s (should be a regular executable file)", name, err)
+	case info.IsDir():
+		return fmt.Errorf("toolchain %s is a directory (should be a regular executable file)", name)
+	case !info.Mode().IsRegular():
+		return fmt.Errorf("toolchain %s is not a regular file (should be a regular executable file)", name)
+	case info.Mode().Perm()&0111 == 0:
+		return fmt.Errorf("toolchain %s is not executable (should be a regular executable file)", name)
+	default:
+		return nil
+	}
 }
 
 // lookInPaths returns all files in paths (a colon-separated list of


cli/units_cmd.go
```diff
@@ -111,6 +119,21 @@ func scanUnitsIntoConfig(cfg *config.Repository, quiet bool) error {
...
 
+func checkRegularExecutableFile(name string) error {
+	switch info, err := os.Stat(name); {
+	case err != nil:
+		return err
+	case info.IsDir():
+		return fmt.Errorf("a directory")
+	case !info.Mode().IsRegular():
+		return fmt.Errorf("not a regular file")
+	case info.Mode().Perm()&0x1 == 0:
```
shurcooL: This is just a minor style suggestions, but unix permissions are usually expressed using octal numbers (e.g., `0777`). So, this could be rewritten to use `01` or `0001`. Do you think that's better than hex or not?

Also, do you want to check just the last bit for "others", what about "owner" and "group" executable bits?

cli/units_cmd.go
```diff
@@ -38,6 +40,12 @@ func scanUnitsIntoConfig(cfg *config.Repository, quiet bool) error {
 		if err != nil {
 			return err
 		}
+
+		// Make sure that the file pointed to by cmdName is a regular executable file.
```
shurcooL: It's probably better to move this comment to the declaration of `func checkRegularExecutableFile`, so its behavior is documented. No need for comment here, since the func name is descriptibe and will be documented after that move, so it's easy to figure out what it does.

cli/units_cmd.go
```diff
@@ -38,6 +40,12 @@ func scanUnitsIntoConfig(cfg *config.Repository, quiet bool) error {
...
+		if err := checkRegularExecutableFile(cmdName); err != nil {
+			return fmt.Errorf("%s should be a regular executable file, but it's %s", cmdName, err)
```
shurcooL: Probably better to move the entire error into that func, so this can be just:

```
if err := checkRegularExecutableFile(cmdName); err != nil {
    return err
}
```

shurcooL:
Looking great! Thanks for doing this, it'll be super handy!

LGTM.

cli/units_cmd.go
```diff
@@ -111,6 +118,22 @@ func scanUnitsIntoConfig(cfg *config.Repository, quiet bool) error {
 	return nil
 }
 
+// checkRegularExecutableFile verifies that the file pointed to by name is a regular executable file.
+func checkRegularExecutableFile(name string) error {
+	reason := error(nil)
```
shurcooL: A more idiomatic way to express this is:

```Go
var reason error
```

The zero value for error type is already nil.

(For reference, see http://dave.cheney.net/2014/05/24/on-declaring-variables.)

cli/units_cmd.go
```diff
@@ -111,6 +118,22 @@ func scanUnitsIntoConfig(cfg *config.Repository, quiet bool) error {
...
+	case err != nil:
+		reason = fmt.Errorf("caused %s", err)
+	case info.IsDir():
+		reason = fmt.Errorf("a directory")
+	case !info.Mode().IsRegular():
+		reason = fmt.Errorf("not a regular file")
+	case info.Mode().Perm()&0111 == 0:
+		reason = fmt.Errorf("not an executable file")
+	}
+	return fmt.Errorf("%s should be a regular executable file, but it's %s", name, reason)
```
shurcooL: This is okay, but if it were me, I wouldn't be afraid to copy the `return fmt.Errorf("%s should be a regular executable file`  part 5 times.

When it comes to DRY, it's valuable to avoid duplication of high level ideas and concepts. But duplicating a few lines of code in a small context where it's clear what's going on is very inexpensive.

Also see Rob Pike's "A little copying is better than a little dependency." from https://go-proverbs.github.io/.

mkovacs: Noted for future reference, thanks! :)

shurcooL: That said, your latest refactored version will always return non-nil error, that's probably not intended.

cli/units_cmd.go
```diff
@@ -111,6 +118,24 @@ func scanUnitsIntoConfig(cfg *config.Repository, quiet bool) error {
...
+		return fmt.Errorf("%s should be a regular executable file, but it caused %s", name, err)
+	case info.IsDir():
+		return fmt.Errorf("%s should be a regular executable file, but it's a directory", name)
+	case !info.Mode().IsRegular():
+		return fmt.Errorf("%s should be a regular executable file, but it is not a regular file", name)
+	case info.Mode().Perm()&0111 == 0:
+		return fmt.Errorf("%s should be a regular executable file, but it is not an executable file", name)
+	default:
+		return nil
+	}
```
shurcooL: Looks awesome! 👍

cli/units_cmd.go
```diff
@@ -111,6 +118,24 @@ func scanUnitsIntoConfig(cfg *config.Repository, quiet bool) error {
 	return nil
 }
 
+// checkRegularExecutableFile verifies that the file pointed to by name is a regular executable file.
+func checkRegularExecutableFile(name string) error {
+	switch info, err := os.Stat(name); {
+	case err != nil && os.IsNotExist(err):
```
shurcooL: `err != nil && os.IsNotExist(err)` can be simplified to just `os.IsNotExist(err)` because that func is safe to run on `nil` error, it will return false.

shurcooL:
LGTM again! 🚢

PR #299, opened 2016-06-10
Updated Python toolchain installation

alexsaveliev:
- Python toolchain is no longer based on Go code so there is no need to clone it to `GOPATH` and make symlink in the `SRCLIBPATH`
- Windows support: On Windows there is no `python3.5` so using `python --version` to check installed Python's version

### DIFF
--- a/cli/toolchain_cmd.go
+++ b/cli/toolchain_cmd.go
@@ -9,8 +9,9 @@ import (
 	"os/exec"
 	"path"
 	"path/filepath"
+	"regexp"
 	"runtime"
-
+	"strconv"
 	"strings"
 
 	"github.com/alexsaveliev/go-colorable-wrapper"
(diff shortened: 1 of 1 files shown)


cli/toolchain_cmd.go
```diff
@@ -740,3 +738,45 @@ func toolchainFQN(toolchainName string) string {
...
+	if err != nil {
+		return fmt.Errorf("unable to detect Python version")
+	}
+
+	version := regexp.MustCompile("^Python (\\d+)\\.(\\d+)").FindStringSubmatch(string(out))
+	if version == nil {
+		return fmt.Errorf("unable to detect Python version from version string '%s'", string(out))
+	}
+
+	major, err := strconv.Atoi(version[1])
```
shurcooL: What if `len(version)` is 0 or 1, wouldn't this line panic?

alexsaveliev: It shouldn't. `version` keeps matching string and all the sub-expressions (re above defines 2 of them). If there is no match, `version` will be `nil`

shurcooL: Understood, thanks!

PR #302, opened 2016-07-08
replace github.com/rogpeppe/rog-go/parallel with github.com/neelance/parallel

neelance:
This avoids messing up CPU and memory profiles by not adding its own stack frame.


keegancsmith:
LGTM

btw I just want to re-iterate how awesome I think this change is. Beautiful solution to our problem.

PR #304, opened 2016-08-29
README: fix typos in goxc path

bradleyfalzon:
The developer instructions contains instructions to install `goxc` utility, there's a minor typo in the import path and users should `go get` the path in order to fetch and install, not `go install` which requires the path to exist locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,7 +69,7 @@ instructions.
 Contributors with deploy privileges can update the official binaries
 via these instructions:
 
-1. `go install github.com/laxer/goxc`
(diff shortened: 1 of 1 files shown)


shurcooL:
I can confirm the typo in the import path of README. The Makefile has the correctly spelled version, see:

https://github.com/sourcegraph/srclib/blob/309efcef61f45418c862e70f2f3d7f93cd9f424e/Makefile#L37

The change to `go get -u` makes sense.

LGTM. Thanks!

